PR #4, opened 2025-10-08
Add undo / redo functionality

akre54:
(Migrated from internal). Work in progress

https://github.com/user-attachments/assets/981e5dcb-bfff-447b-9aab-4cfc883b5ca1


akre54:
I want to land this but I think we need to follow Theatre's state split of 1) historical 2) ephemeral 3) ahistorical. Ephemeral becomes historical when the transaction is committed. Ahistorical doesn't have an undo / redo. Historical does.

PR #9, opened 2025-10-08
Collapse compound props

akre54:
This is a nice change but we don't have a way to reposition the handles based on height currently. TODO


akre54:
Now that #37 has landed I feel better about adding this. It doesn't persist across reloads but it's a nice usability touch

chrisgervang:
Yeah its a good start

PR #22, opened 2025-10-08
Add turfjs geo util operators

akre54:
Add some of the most popular / common turfjs operators for ease-of-use

Also update many of the types to accept either an array of js point objects, or geojson. 

https://github.com/user-attachments/assets/9c4c28d2-8db1-4e91-b49e-fd9c3780f14b

### DIFF
--- a/noodles-editor/src/noodles/fields.ts
+++ b/noodles-editor/src/noodles/fields.ts
@@ -1577,4 +1577,230 @@ export class FeatureCollectionField extends Field<z.ZodType<FeatureCollection<Ge
       : turf.featureCollection([])
     super(initialFC, options)
   }
+
+  // Convert FeatureCollection to tabular format (extract properties)
+  toTabular(): TabularData {
+    const features = this.value.features
+    const properties = features.map((f, i) => {
+      const coords = f.geometry.type === 'Point' ? f.geometry.coordinates : null
+      return {
+        ...f.properties,
+        _lng: coords?.[0],
+        _lat: coords?.[1],
+        _geometryType: f.geometry.type,
+        _featureIndex: i,
+      }
+    })
+    return {
+      length: properties.length,
+      properties,
+      attributes: {},
+    }
+  }
+}
+
+// Helper to detect ALL coordinate pairs in an object (e.g., pickup/dropoff)
+export function detectCoordinateSets(
+  obj: Record<string, unknown>
+): Record<string, { lng: string; lat: string }> {
+  if (typeof obj !== 'object' || obj === null) return {}
+
+  const lngPattern = /^(.+?)_?(lng|lon|longitude)$/i
+  const latPattern = /^(.+?)_?(lat|latitude)$/i
+
+  const sets: Record<string, { lng: string; lat: string }> = {}
+  const lngFields: Array<{ key: string; prefix: string }> = []
+  const latFields: Array<{ key: string; prefix: string }> = []
+
+  // Find all lng/lat fields with their prefixes
+  for (const key of Object.keys(obj)) {
+    if (typeof obj[key] === 'number') {
+      const lngMatch = key.match(lngPattern)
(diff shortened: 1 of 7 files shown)


noodles-editor/src/noodles/fields.ts
```diff
@@ -617,7 +618,7 @@ export class Point3DField extends Field<
   }
 }
 
-type Point2DFieldValue = { lng: number; lat: number; [key: string]: unknown } | [number, number]
+type Point2DFieldValue = { lng: number; lat: number;[key: string]: unknown } | [number, number]
```
chrisgervang: Thats a strange lint

noodles-editor/src/noodles/fields.ts
```diff
@@ -1310,3 +1311,220 @@ export class BezierCurveField extends Field<z.ZodType<BezierCurveData>> {
...
+    return value as FeatureCollection
+  }
+
+  // Array of features
+  if (
+    Array.isArray(value) &&
+    value.length > 0 &&
+    typeof value[0] === 'object' &&
+    'type' in value[0] &&
+    value[0].type === 'Feature'
```
chrisgervang: I wonder if the benefits of flat duck typing outweigh the readability of these if statements throughout. Isn't this what zod is typically used for?

One opportunity is when we're validating to a published geojson spec like `isFeature`, `isFeatureCollection`, and `isGeometry`. I'm a bit surprised neither turf nor `geojson` packages include these but its easy to do ourselves.

Something to look at is we check if something is a Feature in four places in two ways in pointsToFeature and pointsToFeatureCollection, and using a different heuristic, in GeometryField and FeatureField.

noodles-editor/src/noodles/fields.ts
```diff
@@ -1310,3 +1311,220 @@ export class BezierCurveField extends Field<z.ZodType<BezierCurveData>> {
...
+  if (!value) return null
+
+  // Already a GeoJSON Feature
+  if (typeof value === 'object' && 'type' in value && value.type === 'Feature') {
+    return value as Feature
+  }
+
+  // Single point as array: [lng, lat]
+  if (Array.isArray(value) && value.length === 2 && typeof value[0] === 'number') {
+    return turf.point([value[0], value[1]], value)
```
chrisgervang: The properties arg is typically an object
```
return turf.point([value[0], value[1]], { value })
```

Good idea to pass the original through.. lets just use a consistent key. Could be `value`, `source`, `input`

noodles-editor/src/noodles/operators.ts
```diff
@@ -4142,6 +4141,354 @@ export class GeoJsonTransformOp extends Operator<GeoJsonTransformOp> {
...
+}
+
+export class BufferOp extends Operator<BufferOp> {
+  static displayName = 'Buffer'
+  static description = 'Create a buffer around a GeoJSON feature'
+  asDownload = () => this.outputData
+  createInputs() {
+    return {
+      feature: new FeatureField(),
+      radius: new NumberField(1, { min: 0, max: 10000, step: 0.1 }),
```
chrisgervang: This doesn't have a consistent hard max.. it'd vary by unit and even then it's not clear what it'd actually be

noodles-editor/src/noodles/operators.ts
```diff
@@ -4142,6 +4141,354 @@ export class GeoJsonTransformOp extends Operator<GeoJsonTransformOp> {
...
+    feature,
+    tolerance,
+    highQuality,
+  }: ExtractProps<typeof this.inputs>): ExtractProps<typeof this.outputs> {
+    const simplified = turf.simplify(feature, { tolerance, highQuality })
+    return { simplified }
+  }
+}
+
+export class LengthOp extends Operator<LengthOp> {
```
chrisgervang: What's the difference between this and DistanceOp?

PR #25, opened 2025-10-10
Add spec for integrating Claude into the app as a sidebar

akre54:
Basic spec for adding Claude to the app as a client-side agent.

We'd bundle the app code, tests, and docs into a single bundle and build on every push. We'd send this asset over the wire when the sidebar is open to prime the LLM's RAG. We also want to be able to integrate in with Chrome's MCP and capture screenshots to assist in creating visualizations and debugging the framework.

### DIFF
--- a/dev-docs/specs/claude-integration-spec/claude-integration-spec.md
+++ b/dev-docs/specs/claude-integration-spec/claude-integration-spec.md
@@ -1,0 +1,3701 @@
+# Claude AI Integration Spec
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


chrisgervang:
Is there any difference between this and the spec in #26? I wonder if github supports stacked branches like phab does.. these AI spec flows tend to come in twos. It's easier to review smaller PRs

akre54:
Yeah I started with this and then branched #26 for the implementation. This mentions Claude more while the spec there is more open-ended but otherwise they're similar.

I was hoping to use the spec as an alignment doc but it's pretty huge and the code is already far along so maybe it's best to just move discussion there

Github doesn't support stacked PRs, but you can change the upstream branch, which sorta functions similarly

PR #26, opened 2025-10-10
Claude chat implementation

akre54:
This PR adds a Claude chat to the Noodles Editor. It is aimed at beginner and intermediate users who may be learning both Noodles and geospatial / sql / deck for the first time.

Critical user journeys: 
* Basic plotting on a map ("show me all citibike stations in nyc")
* Updating options and state ("make the circles bigger", "turn them into car models")
* Debugging and diagnosing issues ("why can't I see my models on the map")

Summary from Claude Code:
The implementation includes a comprehensive chat UI with conversation history, a client-side Claude API integration that uses user-provided API keys (no server infrastructure required), an MCP-like tool system that allows Claude to read/modify the project state, read documentation and operator metadata, auto-capture screenshots for visual context, and a context generation pipeline via GitHub Actions that indexes source code, documentation, and examples for Claude's reference. 

Testing:
1. Add your anthropic api key to the env file or set it via the web editor
2. Ask one of the critical user journeys above (e.g. "Create a number operator with value 42", "change the color to red")
3. Check visualization and outputs update correctly. Check history and api modal work
4. Report feedback / update prompt as needed

Screenshot 2025-10-20 at 1 44 18 PM
Screenshot 2025-10-20 at 1 44 02 PM


noodles-editor/README.md
```diff
@@ -66,6 +66,26 @@ The main info relevant to Noodles.gl is the `type` of the [Node object](https://
...
+## Environment Variables
+
+You can optionally set environment variables for the Claude AI chat feature:
+
+1. Copy `.env.example` to `.env.local`:
+   ```bash
+   cp .env.example .env.local
+   ```
+
+2. Add your Claude API key:
```
chrisgervang: Mention how to get the others as well?

akre54: Right now it only supports Claude. Or do you mean like Maptiler, GMaps, etc?

noodles-editor/src/noodles/noodles.tsx
```diff
@@ -615,6 +618,11 @@ export function getNoodles(): Visualization {
     }))
   }, [edges])
 
+  const handleProjectUpdate = useCallback((updatedProject: { nodes: AnyNodeJSON[], edges: ReactFlowEdge<unknown>[] }) => {
```
chrisgervang: Does this update all of the state? What about `opMap`?

noodles-editor/src/noodles/components/menu.tsx
```diff
@@ -679,6 +683,36 @@ export function NoodlesMenubar({
...
+              <Menubar.Item className={s.menubarItem}>Hide TheatreJS</Menubar.Item>
+              <Menubar.Separator className={s.menubarSeparator} />
+              <Menubar.Item className={s.menubarItem}>Toggle Fullscreen</Menubar.Item>
+            </Menubar.Content>
+          </Menubar.Portal>
+        </Menubar.Menu>
+
+        {/* Chat button on the right side */}
+        {setShowChatPanel && (
+          <div className={s.menubarRightSlot}>
```
chrisgervang: Can this layout be done with radix?

noodles-editor/.env.local.example
```diff
@@ -2,3 +2,4 @@ VITE_GOOGLE_MAPS_API_KEY=
 VITE_CESIUM_ACCESS_TOKEN=
 VITE_MAPBOX_ACCESS_TOKEN=
 VITE_MAPTILER_API_KEY=
+VITE_CLAUDE_API_KEY=sk-ant-your-api-key-here # For AI chat feature (optional)
```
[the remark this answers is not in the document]
chrisgervang: I'm fine with it either way but we should be consistent with the other env vars. Kepler probably does this best since the error explains how to fix it: https://github.com/keplergl/kepler.gl/blob/master/webpack/shared-webpack-configuration.js#L32

We could do something simpler by just embedding how to get each token in this file, and whether or not its optional.

Alternatively, if there's more context to share than can fit on a line, add notes to the contributing guide.

.github/workflows/generate-context.yml
chrisgervang: I've usually done one website deployment script to keep the deployment a clean, immutable artifact. Does it need to be its own action?

noodles-editor/scripts/generate-context.ts
```diff
@@ -0,0 +1,375 @@
+#!/usr/bin/env tsx
+// Generate context bundles for Claude AI integration
+//
+// This script generates:
```
chrisgervang: Does claude recommend including an https://llmstxt.org/?

noodles-editor/src/noodles/noodles.tsx
chrisgervang: #43 clarified this.. everything is driven by node and edge changes. opMap is updated by transformGraph

noodles-editor/scripts/generate-context.ts
chrisgervang: Right, I meant not role our own llmtxt. It makes sense to generate our own context for the chat bot

> I wonder if we need some kind of automated list of operators and their inputs / outputs in the doc, or just keep it updated manually.

I prefer the api references docs to be written as markdown (as opposed to build-step code reflection link Sphinx) since it’s very flexible, explicit, and easy to update. LLMs make this so much easier to write and maintain - the claude code reviewer could even be configured to [catch when APIs change](https://github.com/anthropics/claude-code-action/blob/main/docs/solutions.md#documentation-sync-on-api-changes).

noodles-editor/src/timeline-editor.tsx
```diff
@@ -173,6 +177,23 @@ export default function TimelineEditor() {
...
       redraw()
     },
+    onAfterRender: () => {
+      visualization.deckProps?.onAfterRender?.()
+
+      // Track FPS and stats for Claude AI debugging
+      const now = Date.now()
+      const deltaTime = now - lastFrameTimeRef.current
+      lastFrameTimeRef.current = now
+      fpsRef.current = deltaTime > 0 ? Math.round(1000 / deltaTime) : 0
```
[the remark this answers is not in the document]
akre54: It's undefined for me. The FPS widget also is also trying to read that prop and failing. Ideas?

.github/workflows/generate-context.yml
akre54: I guess we could potentially be finer grained with the changes than the website push here. And it's technically a separate job so it's ripe for parallelization. But maybe we could combine them yeah

noodles-editor/scripts/generate-context.ts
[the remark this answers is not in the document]
akre54: Ok would you want it to throw?

.github/workflows/generate-context.yml
chrisgervang: It seems to be a choice between deploy faster or be certain the website and context are in-sync. I think there's more reasons to combine them then to parallelize

noodles-editor/src/timeline-editor.tsx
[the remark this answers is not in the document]
chrisgervang: It works for me in the widget. Hmm, I'm guessing it's that deck isn't up to date. We released 9.2 right before the summit. Let's upgrade to `9.2.2`

noodles-editor/src/noodles/components/menu.tsx
```diff
@@ -679,6 +683,36 @@ export function NoodlesMenubar({
...
+              sideOffset={5}
+              alignOffset={-3}
+            >
+              <Menubar.Item className={s.menubarItem}>Hide Nodes</Menubar.Item>
+              <Menubar.Item className={s.menubarItem}>Hide TheatreJS</Menubar.Item>
+              <Menubar.Separator className={s.menubarSeparator} />
+              <Menubar.Item className={s.menubarItem}>Toggle Fullscreen</Menubar.Item>
+            </Menubar.Content>
+          </Menubar.Portal>
+        </Menubar.Menu>
```
chrisgervang: Looks like the robot added these back. They were recently removed

noodles-editor/src/noodles/components/menu.tsx
akre54: Yeah I mean it's not winning any prizes but it looks like this
Screenshot 2025-10-18 at 11 23 04 AM

noodles-editor/src/timeline-editor.tsx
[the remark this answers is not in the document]
akre54: I just checked and it's not in 9.2.2 either. 

I don't see it when I visit https://deck.gl/. Are you sure it landed?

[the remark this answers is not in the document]
akre54: Hm I checked and it's been [landed for a few years](https://github.com/visgl/deck.gl/blame/5b2325d47e7e8e6dd39d4a297e16b3baa1ed6de9/modules/core/src/lib/deck.ts#L55). So no need to upgrade just for this PR (but will do in a follow-up).

noodles-editor/src/ai-chat/chat-panel.tsx
```diff
@@ -0,0 +1,542 @@
...
+      if (!sanitizedKey || !sanitizedKey.startsWith('sk-ant-') || sanitizedKey.length < 20) {
+        localStorage.removeItem('noodles-claude-api-key')
+        setShowApiKeyModal(true)
+        setContextLoading(false)
+        return
+      }
+
+      // If we had to sanitize, save the clean version back to localStorage
+      if (sanitizedKey !== storedKey && localStorage.getItem('noodles-claude-api-key')) {
+        localStorage.setItem('noodles-claude-api-key', sanitizedKey)
```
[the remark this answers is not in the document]
akre54: Is this really an issue? Only this domain can read from localStorage.

It doesn't store the key from vite in localStorage. It has an explicit path for this. The second thing you're saying is what this PR does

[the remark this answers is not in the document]
akre54: From Claude Code

> Let me look at the PR directly on GitHub to understand the security concern better. Based on what I know about API key storage in browsers: Answer about API Key Encryption: The CodeQL bot is technically correct that storing API keys in localStorage is not ideal, but in this specific context, it's not a significant security issue: Why encryption doesn't help much here:
> Browser context limitation: Any encryption you do in JavaScript would need to store the decryption key somewhere accessible to the same JavaScript code, creating a circular problem
> Same-origin access: localStorage is already protected by same-origin policy - only code from the same domain can access it
> Client-side keys are expected: This is a bring-your-own-key (BYOK) pattern where users provide their personal API keys
> The real security considerations:
> ✅ Good: API calls are made client-side directly to Anthropic (not through your servers)
> ✅ Good: The key never touches Noodles.gl servers
> ✅ Good: Users are warned that it's stored locally
> ⚠️ Risk: XSS vulnerabilities could steal the key from localStorage
> ⚠️ Risk: Browser extensions with broad permissions could access it
> Better alternatives (if you wanted to improve it):
> Session-only storage: Use sessionStorage instead - keys cleared when tab closes
> Encrypted IndexedDB: More secure than localStorage, but still vulnerable to XSS
> Server-side proxy: Store keys server-side (but then you have billing/liability concerns)
> My recommendation: For a BYOK pattern in a web app, localStorage is acceptable and common practice. The current implementation with the warning message is fine. The security risk is minimal compared to the convenience, and users who are security-conscious can:
> Use environment variables during development
> Clear localStorage when done
> Use browser profiles or incognito mode
> If you want to address the CodeQL warning, the simplest improvement would be to add a checkbox option: "Remember my API key" (unchecked by default) and use sessionStorage when unchecked.

.github/workflows/generate-context.yml
akre54: It parallelizes easily and is a separate enough workstream that I think it's fine to keep separate for now. If we find that it's taxing resources or it's substantively similar to the website deployment let's reintegrate.

Theoretically these have different dependencies that trigger an update, but we aren't really doing anything smart in that regard now (nor will we likely have to... just pointing out)

noodles-editor/src/ai-chat/chat-panel.tsx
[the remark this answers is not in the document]
chrisgervang: Gotcha. It’s easy to miss that the VITE env key isn’t persisted because it’s temporarily assigned to storedKey, and the comments make it look like it might be. The sanitization logic added noise too.

I’d like to simplify this with a couple changes:

- Remove the sanitize step - It’s unnecessary here. We can just trim() the key - if it’s invalid, the API call will fail gracefully and prompt the user to fix it.
- Stop persisting anything inside this hook - Storing only needs to happen when the user explicitly saves their key in the modal, not during initialization.

I don't see this as a security risk either, but it does make the behavior unpredictable. The app could silently mutate or overwrite their key without them realizing it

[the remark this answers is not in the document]
chrisgervang: I feel more strongly about the UX than security. If it still makes sense to store something during initialize, that's fine. API key ingestion shouldn't try to be smart on the client and it's safe to assume the deployed key is working (or at least, it's not on the user to fix it).. the best practice to validate a key is to test on the server and prompt again if there's an issue with it rather than regex it.

noodles-editor/src/noodles/components/chat-toggle.module.css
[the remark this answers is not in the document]
chrisgervang: Delete unused file?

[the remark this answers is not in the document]
akre54: But we can combine with chat-panel.module. I think this is a historical artifact

PR #35, opened 2025-10-13
Add BlendingOp and `parameters` to every layer

akre54:
Add the ability to enable layer blending (additive, subtractive) to create cool glowing effects

nyc-taxis-blend-add

I don't love adding *yet another* prop to the Layers, but aside from adding a `PropertiesOp` or similar, I think we need to. 

This is also revealing that we need to fix the handle positioning logic, and maybe that we should disable invalid input states (i.e. the `blendColorSrcFactor` and `blendColorDestFactor` inputs should be hidden or disabled unless the `mode` is `'custom'`).


akre54:
After exploring the architecture more deeply, I wanted to share two additional approaches that could address the scalability concerns:

## Alternative 3: Funnel Layers Through BlendingOp

Instead of adding `parameters` to every layer, BlendingOp could accept layers as input and output blended layers:

```typescript
export class BlendingOp extends Operator<BlendingOp> {
  createInputs() {
    return {
      layers: new ListField(new LayerField()), // Accept multiple layers
      mode: new StringLiteralField('normal', {
        values: ['normal', 'additive', 'subtractive', 'custom'],
      }),
      // ... blend factor fields
    }
  }
  createOutputs() {
    return {
      layers: new ListField(new LayerField()), // Output blended layers
      parameters: new UnknownField({}), // Keep for backward compat
    }
  }
  execute({ layers, mode, ... }) {
    const params = this.computeBlendParameters(mode, ...)
    const blendedLayers = layers.map(layer => ({
      ...layer,
      parameters: { ...(layer.parameters || {}), ...params }
    }))
    return { layers: blendedLayers, parameters: params }
  }
}
```

**Graph pattern:**
```
/pickup-layer → /blending.par.layers
/arc-layer → /blending.par.layers
/dropoff-layer → /blending.par.layers
/blending.out.layers → /deck.par.layers
```

**Pros:**
- Scales automatically - add any number of layers to the ListField
- Single BlendingOp node applies to all connected layers
- Most "node-editor-native" approach (similar to GeoNodes/TouchDesigner)
- No need to add `parameters` field to every layer operator

**Cons:**
- Breaking change to BlendingOp signature
- Can't selectively blend some layers in a group (all get same blending)
- Medium implementation effort

## Alternative 4: Built-in Blend Field (like visibility)

Add `blendingMode` as a built-in field to every layer operator, similar to how `visible` works:

```typescript
export class ScatterplotLayerOp extends Operator<ScatterplotLayerOp> {
  createInputs() {
    return {
      visible: new BooleanField(true),
      blendingMode: new StringLiteralField('normal', {
        values: ['normal', 'additive', 'subtractive'],
        showByDefault: false,
      }),
      // ... other fields
    }
  }
}
```

Then modify `parseLayerProps()` to convert `blendingMode` into the appropriate `parameters` object.

**Pros:**
- Best end-user experience - just set a field, no extra nodes
- Consistent with existing `visible` pattern
- Most discoverable in the UI
- Can still connect BlendingOp output to control multiple layers

**Cons:**
- Requires modifying ~30 layer operators
- Adds field clutter even when hidden by default
- Less composable than operator-based approach
- Harder to create reusable blending templates

---

## Comparison

| Approach | Code Changes | Scalability | Composability | UX |
|----------|--------------|-------------|---------------|-----|
| Current (parameters field) | ~30 operators | Manual per-layer | Medium | OK |
| Alternative 3 (Funnel) | BlendingOp only | Automatic | High | Very Good |
| Alternative 4 (Built-in) | ~30 operators + parseLayerProps | Automatic | Low | Best |

I'm leaning toward **Alternative 3** as it provides the best balance of scalability, composability, and implementation effort. It also aligns with how other node editors handle "apply effect to multiple objects" patterns.

Thoughts?

PR #36, opened 2025-10-13
Doc updates and first project walkthrough

akre54:
Add a doc for animation and rendering, and a walkthrough video to get started quickly. Also add a doc of prior art for node-based systems. Helpful for learning the background of the tool and other places we can draw inspiration from.


chrisgervang:
It's a good question.. I'd like to do some research and try different approaches. Caching and dependency installation in GH actions has always been slow and problematic for a long tail of changing reasons. 

Lately I've seen devs move away from Volta to NodeJS's own [corepack](https://github.com/nodejs/corepack) and I'm curious if that project has addressed any underlying issues.

vis.gl projects also take about a minute to install deps for a similar reason. In the past, ocular (deck's build tool) did a hack around where puppeteer was installed globally, cached separately, and then linked into the project deps.. but we moved away from that since it was causing some people issues a couple years ago.

PR #37, opened 2025-10-13
Fix for handle heights to be based on css rather than calculating based on index

akre54:
Puts handles inside the FieldComponent instead of on the OpComponent level


chrisgervang:
> I'm cool with that. Want to make the change?

Happy to but wont realistically be able to this week. Feel free to do it if you need it sooner

PR #47, opened 2025-10-21
Add a watcher plugin to vite for context changes

akre54:
We currently only automatically generate context changes on push to github. This allows localhost to be kept up to date on any changes, for developing with the agent locally


chrisgervang (approved):
Are the artifacts gitignored?

Seems fine, main concern here would be if it's an expensive script and if these files change a lot during, say, an agent generation session

PR #48, opened 2025-10-21
Add CUJs, READMEs for examples, and attempt to fix context

akre54:
Add new MCPs to handle node and edge modification, and extract out the hooks from the noodles.tsx react component, both for testability and to share between the MCP and the UI.

Screenshot 2025-10-21 at 1 38 36 PM
Screenshot 2025-10-21 at 1 39 20 PM

With own-properties:
Screenshot 2025-10-21 at 1 46 14 PM

### DIFF
--- a/noodles-editor/src/noodles/hooks/use-project-modifications.ts
+++ b/noodles-editor/src/noodles/hooks/use-project-modifications.ts
@@ -72,15 +72,20 @@ export function useProjectModifications(options: UseProjectModificationsOptions)
     [setNodes, setEdges]
   )
 
-  // Delete nodes with intelligent edge handling (same logic as noodles.tsx onNodesDelete)
-  const deleteNodes = useCallback(
-    (nodeIds: string[]): ModificationResult => {
-      const nodes = getNodes()
+  // Handle edge reconnection after nodes are deleted (same logic as noodles.tsx onNodesDelete)
+  // This is a callback invoked AFTER nodes have been deleted by ReactFlow
+  // Takes the deleted nodes directly (not IDs) since they're no longer in the nodes array
+  const handleNodesDeleted = useCallback(
+    (deletedNodes: ReactFlowNode[]): ModificationResult => {
+      // Get the current state BEFORE the nodes were deleted (for edge reconnection logic)
+      // We need to reconstruct the pre-deletion state by adding deleted nodes back
+      const currentNodes = getNodes()
+      const nodes = [...currentNodes, ...deletedNodes]
       const edges = getEdges()
-      const nodesToDelete = nodes.filter(n => nodeIds.includes(n.id))
+      const nodesToDelete = deletedNodes
 
       if (nodesToDelete.length === 0) {
-        return { success: false, error: `No nodes found with IDs: ${nodeIds.join(', ')}` }
+        return { success: false, error: `No nodes provided for deletion` }
       }
 
       const warnings: string[] = []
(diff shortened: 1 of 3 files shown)


noodles-editor/src/noodles/hooks/use-project-modifications.ts
```diff
@@ -0,0 +1,748 @@
...
+
+      return { success: true }
+    },
+    [getNodes, setNodes]
+  )
+
+  // Apply a batch of modifications atomically
+  // All nodes are added first, then edges are validated and added
+  // Edge validation failures are logged but don't stop other edges from being applied
+  const applyModifications = useCallback(
```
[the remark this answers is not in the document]
chrisgervang: Can we split this hook change out from the AI chat changes? It's difficult to review

noodles-editor/public/noodles/california-earthquakes/README.md
chrisgervang: Can we split these markdown files out to review them separately?

noodles-editor/src/ai-chat/claude-client.ts
```diff
@@ -57,21 +57,23 @@ export class ClaudeClient {
...
-        screenshot = result.data.screenshot
-        screenshotFormat = result.data.format || 'jpeg'
-      }
-    }
+    // Auto-capture is disabled by default - too large for context
+    // AI should explicitly use capture_visualization tool when needed
+    const screenshot = params.screenshot
+    const screenshotFormat = params.screenshotFormat || 'jpeg'
+
+    // Disable auto-capture to reduce token usage
```
chrisgervang: Delete this auto-capture code, or is this temporary?

noodles-editor/src/ai-chat/claude-client.ts
```diff
@@ -251,39 +260,209 @@ export class ClaudeClient {
...
     return {
       message: finalText,
-      projectModifications,
+      projectModifications: allModifications,
       toolCalls
     }
   }
 
-  private buildSystemPrompt(project: any): string {
+  buildSystemPrompt(project: any): string {
```
chrisgervang: There are a lot of escaped chars in this string and it's only using JavaScript for the first line. It'd be easier to edit and preview this if it were in its own markdown file, and then pre-pend the first line at runtime. Vite can load most file types in as an import too

akre54: Yeah that's a valid point. I added MCPs for grabbing the project state so it really shouldn't need this. Will update

noodles-editor/src/noodles/noodles.tsx
```diff
@@ -168,191 +162,30 @@ export function getNoodles(): Visualization {
...
+    globalContextManager.startLoading().catch(error => {
+      console.warn('Failed to preload AI context:', error)
+    })
+  }, [])
+
   // `transformGraph` needs all nodes to build the opMap and resolve connections
   const operators = useMemo(() => transformGraph({ nodes, edges }), [nodes, edges])
 
+  // Use shared hook for project modifications
+  const { onConnect, onNodesDelete } = useProjectModifications({
```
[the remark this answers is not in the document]
chrisgervang: Yeah that makes sense. I agree the risk is low in principle since it’s just the same logic in a new place. If we already had tests for this I wouldn’t have mentioned it, but since we don’t, it feels like a good opportunity to add some and make sure behavior stays consistent.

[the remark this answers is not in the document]
akre54: Yep - and I don't think it's a bad practice to try to add some tests for this kind of thing. I just want to be thoughtful about where we add process and overhead, because there are *lots* of places that can slow our development down

PR #55, opened 2025-10-22
Move example jsons to subfolders

akre54:
This PR normalizes all example projects to be in a subfolder of `public/noodles/`, to normalize their directory structure and make it easier for the folder loading logic to work.

@chrisgervang what do you think about renaming `/public/noodles` to `/public/examples` instead so we don't have `public/noodles/new/noodles.json`?


noodles-editor/src/noodles/components/project-not-found-dialog.tsx
```diff
@@ -2,7 +2,7 @@ import * as Dialog from '@radix-ui/react-dialog'
 import { Cross2Icon } from '@radix-ui/react-icons'
 import cx from 'classnames'
 import { useCallback, useState } from 'react'
-import newProjectJSON from '../../../public/noodles/new.json'
+import newProjectJSON from '../../../public/noodles/new/noodles.json'
```
akre54: Vite doesn't like that we're importing from the public folder. This should probably be a `fetch`

akre54: #56

PR #60, opened 2025-10-22
Add AGENTS.md and llms.txt

akre54:
Add some context for the LLMs to be able to understand the project without crawling through folders or docs pages.

From what I'm reading these are nascent standards and aren't widely adopted but it can't hurt to add. Maybe in the future we can build them automatically from the existing docs, similar to how we're building the context for the LLM chat agent.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -523,6 +523,88 @@ try {
 - Keep documentation up-to-date
 - Test operators in isolation
 
+## Testing Strategy
+
+### When to Add Tests
+
+**Always add tests for:**
+- New operators and core functionality
+- Changes to critical components (listed below)
+- Complex state management or hook modifications
+- Bug fixes to prevent regressions
+- Non-trivial utility functions
+
+**Test Types:**
+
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,723 @@
...
+- Complex state management or hook modifications
+- Bug fixes to prevent regressions
+- Non-trivial utility functions
+
+**Test Types:**
+
+- **Unit Tests**: For operator logic, pure functions, and utilities
+- **Integration Tests**: For graph transformations, hook interactions, and data flow
+- **Component Tests**: For React components with React Testing Library
+- **E2E Tests**: For full user workflows with Playwright
```
chrisgervang: Our current e2e test doesn't use playwright.. does it? Should it?

Could be good to add some guidance like we do for integration below such as

E2E

- prove the critical journeys work from “load project” to “inspect final flow output.”
- use Playwright sparingly for the most critical user journeys. Mouse/keyboard driven tests are inherently brittle. Prefer unit/integration/component tests for depth and coverage.

PR #62, opened 2025-10-22
Add TimeSeriesOp for time-based data interpolation

chrisgervang:
Introduces the TimeSeriesOp operator to interpolate time series data at a given time using linear interpolation. Includes comprehensive tests for edge cases, field passthrough, and correct interpolation of numeric values.

The API aligns closely to TripsLayerOp, and it's expected they'll typically be used together.

Screenshot 2025-11-11 at 11 42 18 AM

https://github.com/user-attachments/assets/c0c98e82-d40f-438d-83dc-dd36818621b9

### DIFF
--- a/noodles-editor/public/noodles/world-flights/noodles.json
+++ b/noodles-editor/public/noodles/world-flights/noodles.json
@@ -8,8 +8,8 @@
         "locked": false
       },
       "position": {
-        "x": 1200,
-        "y": 150
+        "x": 1875.0367256829527,
+        "y": 204.36993456417756
       }
     },
     {
@@ -20,8 +20,8 @@
         "locked": false
       },
       "position": {
-        "x": 1500,
-        "y": 150
+        "x": 2175.0367256829527,
+        "y": 204.36993456417756
       }
     },
     {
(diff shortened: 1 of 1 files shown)


noodles-editor/src/noodles/operators.ts
```diff
@@ -5014,6 +5014,136 @@ export class MaskExtensionOp extends Operator<MaskExtensionOp> {
...
+    return { time: currentTime }
+  }
+
+  // If only one data point, return it
+  if (timeSeries.length === 1) {
+    return { ...timeSeries[0] }
+  }
+
+  // Sort by time to ensure proper interpolation
+  const sortedData = [...timeSeries].sort((a, b) => a.time - b.time)
```
akre54: Is there a way that we can assume this data is sorted, or give you an option to sort it ahead of time? That way we can use a binary search which will be much faster

noodles-editor/src/noodles/operators.ts
```diff
@@ -5014,6 +5014,136 @@ export class MaskExtensionOp extends Operator<MaskExtensionOp> {
...
+  // If currentTime is after last point, return last point
+  if (currentTime >= sortedData[sortedData.length - 1].time) {
+    return { ...sortedData[sortedData.length - 1] }
+  }
+
+  // Find the two points to interpolate between
+  let beforeIndex = 0
+  let afterIndex = 1
+
+  for (let i = 0; i < sortedData.length - 1; i++) {
```
akre54: Can we do a binary search since we know this is already sorted?

PR #63, opened 2025-10-22
Add Temporal types for DateFields

akre54:
Running into issues around js dates and local timezones. Temporal is already in the framework and allows us to do things like date math or intervals

Runbook: Create a DateOp and view it with a Viewer

### DIFF
--- a/docs/developers/contributing.md
+++ b/docs/developers/contributing.md
@@ -1,0 +1,1 @@
+../../CONTRIBUTING.md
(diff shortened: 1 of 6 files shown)


chrisgervang (commented):
What's a good run book we can both use to test that this works in both of our time zones?

akre54:
I guess simultaneously open a project and check that we're seeing the same values? 

What would you want to see beyond the test cases?

chrisgervang:
Old:

Screenshot 2025-10-26 at 9 07 07 PM

New behavior: 

Screenshot 2025-10-26 at 9 04 57 PM

PR #64, opened 2025-10-23
Allow the Add Node Menu to search by descriptions too

akre54:
To increase discoverability in the operators search box, we should use the descriptions as part of the search options.

Run book:
* Open the add node menu with right click / 'a' key, or the block library with tab and search for "json" or "fetch json" - it should now include the FileOp. 

Maybe in the future we can do a bit more with natural language or something.

### DIFF
--- a/noodles-editor/src/noodles/components/block-library.tsx
+++ b/noodles-editor/src/noodles/components/block-library.tsx
@@ -26,220 +26,220 @@ type BlockLibraryProps = {
 }
 
 export const BlockLibrary = forwardRef<BlockLibraryRef, BlockLibraryProps>(({ reactFlowRef }, ref) => {
-    const { addNodes, addEdges, screenToFlowPosition } = useReactFlow()
-    const [isOpen, setIsOpen] = useState(false)
-    const { currentContainerId } = useSlice(state => state.nesting)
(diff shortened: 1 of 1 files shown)


akre54:
Well it's sorted secondarily by category, not by match.

I feel like it could go either way but it might be confusing to reorder the ops?

chrisgervang:
Ah, I think the category sort might be throwing it off. I made a project tonight searching for exact matches and the menu placed many ops low in my results, sometimes below the fold. Number and Round were tough to find.

I think it's worth looking at this again and streamlining the behavior. I don't find it confusing for the order to change in a search - in fact, I'd expect it.

When thinking about the node menu users, I see a familiar spectrum similar to other search/discovery experience:

- 1. "I have no idea what I'm looking for"
  - categories and sidebar are a good fit - the North Star would be op recommendations where the app knows what the user wanted
- 2. "I roughly know what I want"
  - fuzzy search and refinement filters like "for objects", "for arrays", etc would be useful
- 3. "I know exactly what I want"
  - they're searching for exact matches

For 2) and 3), I’d consider a different layout for search results - not sorted by category, but instead ranked by name/description relevancy. A simple vertical list could work, or a space-efficient wrapping left to right list layout

PR #71, opened 2025-10-23
Fix yarn cache github action

akre54:
Github push is failing on the `Post Setup Node.js` step, likely due to the fact that our yarn files aren't in the root directory but rather subdirectories. This should fix the issue with caching.


chrisgervang:
Actions and cacheing are quite finicky. Thanks for tackling this

PR #79, opened 2025-10-24
Attempt to fix playwright test again. 🤞

akre54:
Last fix failed. Hopefully this one gets it


akre54:
Oh man this runs so much faster. From like 6 minutes down to 48 seconds

charlieforward9:
Ive been battling these issues as well in a seperate project, thankful to be following this along to see how to speed things up 🤝

chrisgervang:
That's a bummer. Sounds like it'd be worth switching our reviewer to Claude then 😅 most of Copilot's suggestions have been just noise

PR #80, opened 2025-10-24
Fix for geocoder double input render - now adds a renderInput prop on FieldComponent

akre54:
This broke in #37 because it didn't take into account the Geocoder edge case. It's the only operator with both inputs and a custom field component (the other custom op components like MouseOp, TimeOp, etc, don't have inputs. ViewerOp doesn't have any FieldComponents). 

Prev:

Screenshot 2025-10-24 at 3 00 22 PM

New:

Screenshot 2025-10-24 at 3 02 53 PM


akre54:
Nope! Because that's a separate Field type.

PR #84, opened 2025-10-24
Add back in React Strict Mode and fix Theatre prism warnings

akre54:
Fix the theatre prism warnings and turns React Strictmode back on. It fixes the prism warnings by moving the theatre transaction logic to the end, after all of the computation has happened. This makes it a bit clearer and more straightforward

This adds another sheet to the sidebar, so we should figure out the project loading logic. Suggestions @chrisgervang? 

Runbook: Change values in the Nodes UI and set keyframes in Theatre UI to see if they update correctly or throw an error


chrisgervang:
Strict mode calls all hooks twice, including ones that interface with storage APIs. We should carefully run through every core user flows before switching it back on, and maybe ask the robot to review our code's readiness for strict mode. 

Maybe split the theatre change out if it's useful on its own?

PR #95, opened 2025-10-28
Add an Import button to the Not Found project modal

akre54:
Before:

Screenshot 2025-10-27 at 8 57 10 PM

After:

Screenshot 2025-10-27 at 8 54 13 PM

Runbook: Visit `?project=foo` or any missing project and then click the `import project` button. You should be able to load one from disk that mirrors the export flow

Alternatively we could update the export flow to download a zip file with all assets, similar to Theatre, and preserving the projectName and noodles.json convention that our imports have


chrisgervang:
Just tested by importing an old project and it seems to work.

> I kind of like that the json format is open and hackable

Me too. I like the idea of giving the user a choice to export with or without the data, and giving them an easy import for both as well

PR #96, opened 2025-10-28
Export project as zip

akre54:
To match our current directoryHandle structure, we should really be exporting the whole project directory, including data and a root noodles.json file, instead of a json file named the same as the project as we do currently.

Runbook: Visit any project (say `?project=nyc-taxis`) and go to File > Download Project. It should download an `nyc-taxis.zip` file. When you unzip the file you should see a folder called `nyc-taxis`. 

Alternative to #95


akre54:
Because it's a big dependency and it's only used for this specific use
case, so we lazy load to keep the core small

PR #97, opened 2025-10-28
Fix us employment example which was invisible because of a missing value

akre54:
Before:
Screenshot 2025-10-27 at 9 46 58 PM

After:
Screenshot 2025-10-27 at 9 46 11 PM

Map zoom is still offset, but that's something different

We also should be friendlier about surfacing these types of errors. They're easy to fix but frustrating to find if you're not familiar.

### DIFF
--- a/noodles-editor/public/noodles/california-earthquakes/noodles.json
+++ b/noodles-editor/public/noodles/california-earthquakes/noodles.json
@@ -74,7 +74,9 @@
       "id": "/magnitude-color",
       "type": "ColorRampOp",
       "data": {
-        "inputs": {},
+        "inputs": {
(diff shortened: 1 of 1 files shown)


chrisgervang (approved):
Oof, nice find. It's like a needle in a hay stack. Any ideas on how we can surface these? Would it help to add an "undefined value returned" detector to accessors?

I can't think of a case where you'd _want_ an accessor's value to return undefined. Can you?

akre54:
We could have the Accessor set the `executionState` of the operator directly, though that's a little messy. Or have some way of passing back a callback signal from the execute function. Maybe we can add a second options arg to execute?

PR #98, opened 2025-10-28
Add a MapLibreBaseMapOp to new project

akre54:
I find this is something I do for most projects, so it would make sense to include it in the "base" project.

I still find it confusing that there is a ViewState and a Basemap op that do similar things but are slightly different. Remind me why these are here and different @chrisgervang? 

My preference would be to reach a future where these inputs are parameters in a side window for the DeckRendererOp instead of draggable inputs as a separate Op, mostly for streamlining sake. For now this'll do.

Before:
Screenshot 2025-10-27 at 9 52 18 PM

After:
Screenshot 2025-10-27 at 9 51 53 PM

### DIFF
--- a/noodles-editor/public/noodles/new/noodles.json
+++ b/noodles-editor/public/noodles/new/noodles.json
@@ -23,6 +23,18 @@
         "x": 390,
         "y": 0
       }
+    },
+    {
+      "type": "MaplibreBasemapOp",
+      "id": "/maplibre-basemap",
+      "position": {
+        "x": -27.097062041233293,
+        "y": 282.8294486649364
+      },
+      "data": {
+        "inputs": {},
+        "locked": false
+      }
     }
   ],
   "edges": [
@@ -32,10 +44,20 @@
       "target": "/out",
       "sourceHandle": "out.vis",
       "targetHandle": "par.vis"
+    },
+    {
+      "source": "/maplibre-basemap",
+      "sourceHandle": "out.maplibre",
+      "target": "/deck",
+      "targetHandle": "par.basemap",
+      "id": "/maplibre-basemap.out.maplibre->/deck.par.basemap"
     }
   ],
-  "viewport": {},
-  "version": 6,
+  "viewport": {
+    "x": 487.61570812099455,
+    "y": 199.8206070154179,
+    "zoom": 0.7608097176595698
+  },
   "timeline": {
     "sheetsById": {
       "Noodles": {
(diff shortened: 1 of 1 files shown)


chrisgervang:
> I still find it confusing that there is a ViewState and a Basemap op that do similar things but are slightly different. Remind me why these are here and different @chrisgervang?

`viewState` could possibly be removed in the future, but it is a core api so I imagine deck users will be surprised its missing. A reason it can be removed is the `views: View[]` ops all also have a viewState field corresponding to their view, and it's a simplier API (e.g. the top level one supports nested state keyed on viewId)

The maplibre/`basemap` field both configures the map style and constrains viewState to the format compatible with maplibre (like pitch max: 60), and its an implicit configuration flag too: deck+maplibre render when it's defined, or just deck renders when its not defined. There are meaningful differences between the two modes.. interactivity, how views work, and the beforeId layer props are a few (see MapboxOverlay [limitations](https://deck.gl/docs/api-reference/mapbox/overview#limitations)).

More background in https://linear.app/joby/issue/VIS-905/deckrendererop-refactor and https://github.com/joby/visualization/pull/610

----

FWIW, I don't think this API is finished. I'll surface some of our earlier assumptions and open questions:

## combined renderer op vs op-per-"mode"

At the time we explored some high-level alternatives like making a RendererOp per rendering "mode" (i.e. deck, deck+maplibre, future deck+googlemaps, etc) - but we preferred fewer renderer nodes, so the direction moved towards adding a `basemap` field instead.

If we instead had a `MaplibreRendererOp`, it's fields could be the same except we'd replace `basemap` with `mapStyle` - while the APIs nearly the same today, props behave differently (notably, `views` prop [remarks](https://deck.gl/docs/api-reference/mapbox/overview#limitations) and the limitations linked above).

The deck vs deck+maplibre APIs could diverge further if we choose to support more [maplibre-specific props](https://visgl.github.io/react-map-gl/docs/api-reference/maplibre/map). [Controls](https://visgl.github.io/react-map-gl/examples/maplibre/controls) (like deck widgets, but for maplibre), sky, light, and a long tail of camera options (not that all should be top-level fields)

Conclusion was: Adding a `basemap` field to DeckRendererOp was seen as better than having two RendererOps. All of these maplibre fields could be added to the MaplibreBasemapOp, so it made sense to keep the "combined" op.

## Open questions

- How might the framework help switch between desired modes (deck or deck+maplibre) without the user needing to wire up a null op manually?
- How might we support googlemaps basemaps that support also interleaved rendering? GoogleMapsBasemapOp?
- How might we make the differences between rendering modes easier to learn? (API changes, docs, examples, etc)

PR #106, opened 2025-10-29
Add wrangler config for cloudflare pages to pick up

akre54:
Testing this out. Ideally it should pick up from the wrangler config


akre54:
Oh apparently you have to set these all through the dashboard, not in the repo

Issue #131, opened 2025-11-12
[Bug] Sheet object naming collisions in containers


akre54:
### Description

Sheet object names are [keyed by their basename](https://github.com/joby-aviation/noodles.gl/blob/847477e6fc078c0080ec9f0ccc40d55954d4601f/noodles-editor/src/noodles/theatre-bindings.ts#L170) in the theatre binding in order to keep the sidebar clean.

They're keyed by their full id in the sheetObjectMap which keeps them globally unique. But it also means that you can't have two same-named operators at any point in the hierarchy. This isn't an issue in the same level because we have naming collision detection.

Maybe a solution is to use the id without the leading slash for the sheet object name - so `/multiply` becomes just `multiply`, but `/container-1/multiply` becomes `container-1/multiply`. Or we can split on the slash and [namespace the object](https://www.theatrejs.com/docs/latest/manual/objects#namespacing-objects).

### Expected Behavior

I'd expect to be able to create a `multiply` node at the root level, and then another one in a child container without the app crashing.

### Steps to Reproduce

1. Create a `multiply` node at the root level
2. Create a container and dive into the container
3. Create a `multiply` node. The app crashes

### Environment

- Noodles version: 0.1.0
- Browser:
- OS:

akre54:
@copilot update `bindOperatorToTheatre` in theatre-bindings.ts to change the sheet key from `getBaseName(op.id)` to `op.id.slice(1).split('/').join(' / ')`. Make sure to clean up the `getBaseName` import and make sure the tests pass

akre54:
@copilot

PR #132, opened 2025-11-12
Move away from global opMap store to use zustand instead

akre54:
Ideally this is no visible change from the outside, but it'll fix a lot of the reactive bugs we see where state doesn't update. Also it should make it easier to test and mock

### DIFF
--- a/noodles-editor/src/noodles/__tests__/node-operations-integration.test.tsx
+++ b/noodles-editor/src/noodles/__tests__/node-operations-integration.test.tsx
@@ -4,17 +4,16 @@ import type { Edge as ReactFlowEdge, Node as ReactFlowNode } from '@xyflow/react
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import {
   clearOps,
-  deleteOp,
-  getAllOps,
   getOp,
   getOpStore,
   hasOp,
-  setOp,
 } from '../store'
 import { transformGraph } from '../transform-graph'
 import { edgeId, nodeId } from '../utils/id-utils'
 import { getBaseName } from '../utils/path-utils'
 import { serializeNodes } from '../utils/serialization'
+// Import operators to ensure they're registered before tests run
+import '../operators'
 
 // Mock Theatre.js studio to avoid side effects
 vi.mock('@theatre/studio', () => ({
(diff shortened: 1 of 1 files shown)


noodles-editor/src/noodles/components/op-components.tsx
```diff
@@ -512,32 +513,33 @@ function NodeHeader({ id, type, op }: { id: string; type: OpType; op: OperatorIn
...
+          queueMicrotask(() => store.deleteOp(oldChildId))
         }
       }
 
       // Give React time to update the component tree before deleting the old id
       // TODO: this is a hack. We should hook into some sort of "after update" event
       queueMicrotask(() => {
-        opMap.delete(id)
+        store.deleteOp(id)
       })
```
chrisgervang: I wonder if this is still needed? Note to look back at what issue this micro task solved

noodles-editor/src/noodles/store.tsx
```diff
@@ -1,91 +1,142 @@
...
 
-export const useNestingStore = create<NestingState>((set) => ({
-  currentContainerId: '/',
-  setCurrentContainerId: (id: string) => set({ currentContainerId: id }),
-}))
+export const useOperatorStore = create<OperatorStore>((set, get) => ({
+  operators: new Map(),
+  sheetObjects: new Map(),
+  hoveredOutputHandle: null,
+  version: 0,
```
chrisgervang: What would version be useful for?

akre54: I think this was more to do with adding a history entry, but then it's not as important now. I can remove

akre54:
Ok added automated integration tests. They're not perfect because a lot of it relies on race conditions or real timers, but they're there.

I tested the cases you brought up. Copy and paste seems to work, as does copying and pasting many nodes with edges.

Pasting into containers is broken, I think because of #131. Renaming nodes with upstream inputs is broken but that existed before too (#133). 

That is to say, I don't think this PR broke anything new as far as I can tell (famous last words...). Hopefully those other two are easy to test and fix and then rule out

AGENTS.md
```diff
@@ -719,8 +817,11 @@ When implementing features or fixes:
...
 10. **Keep PRs focused** - Split large changes into reviewable chunks when possible
+11. **Use store helpers** - Always use `getOp()`, `getAllOps()`, etc. instead of direct `opMap` access
+12. **Batch store updates** - Use `getOpStore().batch()` when making multiple related changes
+13. **Non-reactive store access** - Store helpers don't trigger React re-renders; use hooks for reactive access
 
 ---
 
-**Last Updated**: 2025-10-22
-**Version**: Based on project version 6 schema
+**Last Updated**: 2025-01-12
```
akre54: Wrong date

PR #134, opened 2025-11-12
Fix Theatre naming collision with containers

### LINKED ISSUE #131: [Bug] Sheet object naming collisions in containers

akre54:
image

Fixes crash where conflicting names lead to a crash. We should now be able to copy containers with no issues.

Closes #131

### DIFF
--- a/noodles-editor/src/noodles/__tests__/theatre-bindings-integration.test.tsx
+++ b/noodles-editor/src/noodles/__tests__/theatre-bindings-integration.test.tsx
@@ -334,4 +334,84 @@ describe('Theatre bindings integration', () => {
       cleanup()
     }
   })
+
+  it('should handle naming collisions for operators with same base name at different hierarchy levels', () => {
+    // This test verifies the fix for issue #131
+    // Operators with the same base name at different levels should not collide
+    const nodes = [
+      {
+        id: '/number',
+        type: 'NumberOp',
+        position: { x: 0, y: 0 },
+        data: { inputs: { value: 2 } },
+      },
+      {
+        id: '/container',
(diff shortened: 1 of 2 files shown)


chrisgervang (approved):
Nice, simple fix!

chrisgervang:
> We should now be able to copy containers with no issues.

Yeah, theatre paths were the last part of _copying_ that seemed broken. I think pasting is another story though and could still have issues until the state systems between react-flow and ops are unified. I'll test pasting out today

PR #137, opened 2025-11-12
Make ContainerOp children count reactive

akre54:
Previously the counter showing the number of children didn't update when a user creates ops inside of a container

This was a regression. It initially worked.

After:
Screenshot 2025-11-12 at 1 56 40 PM


noodles-editor/src/noodles/components/op-components.tsx
```diff
@@ -1120,8 +1119,13 @@ function ContainerOpComponent({
   }
 
   const setCurrentContainerId = useNestingStore(state => state.setCurrentContainerId)
-  const nodes = useNodes()
```
chrisgervang: Odd regression considering this is a reactive hook (was looking at their [docs](https://reactflow.dev/api-reference/hooks/use-nodes), since other hooks of theirs aren't reactive.)

Are nested nodes in react-flow, but hidden? Or maybe it changes, but not when the container is mounted?

PR #140, opened 2025-11-12
[fix] shouldn't undo graph changes when timeline is active

chrisgervang:
Runbook:
- open a project
- add a node
- open timeline and add a keyframe
- move the keyframe
- ctrl+z

It should only undo the keyframe move. The added node shouldn't be undone.

---

Modified noodles-editor/src/noodles/components/UndoRedoHandler.tsx:38-88 to add context-aware undo/redo handling:

The Fix:
  - Added detection for when Theatre.js timeline is focused by checking if the event target is within Theatre.js shadow DOM (#theatrejs-studio-root)
  - When Theatre.js is focused, the keyboard handler now skips graph undo/redo operations and lets Theatre.js handle its own undo/redo
  - The check uses composedPath() to detect events originating from within the shadow DOM

How it works:
  1. When Cmd+Z or Cmd+Shift+Z is pressed, the handler first checks if Theatre.js has focus
  2. If Theatre.js is focused (user is working in the timeline), the graph undo/redo is skipped
  3. If Theatre.js is NOT focused (user is working in the graph), the graph undo/redo proceeds normally

This ensures that when you're working on keyframes in the Theatre.js timeline, only the timeline changes are undone/redone, and the flow graph remains unchanged. When working on the graph itself, only graph changes are affected.

### DIFF
--- a/noodles-editor/src/noodles/components/UndoRedoHandler.tsx
+++ b/noodles-editor/src/noodles/components/UndoRedoHandler.tsx
@@ -36,6 +36,39 @@ export const UndoRedoHandler = forwardRef<UndoRedoHandlerRef>((_, ref) => {
   // Add keyboard shortcuts
   useEffect(() => {
     const handleKeyDown = (e: KeyboardEvent) => {
+      // Check if Theatre.js timeline is focused
+      // Theatre.js uses a shadow DOM with id 'theatrejs-studio-root'
+      // If the event target or active element is within Theatre's shadow DOM,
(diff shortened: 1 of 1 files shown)


akre54:
Yeah cool by me but maybe it's more about stopping propagation / bubbling from `theatreRoot`?

chrisgervang:
Yeah good idea, I'll look to see if there's a more wholistic way to implement this. We don't want any cross talk between theatre and our UI.. copy/paste/undo/redo and even spacebar (which is what my other PR is about).

PR #142, opened 2025-11-14
[feat] Add second and millisecond precision to DateFieldComponent

chrisgervang:
Screenshot 2025-11-14 at 2 50 35 PM

### DIFF
--- a/noodles-editor/src/noodles/__migrations__/008-rename-date-to-datetime.test.ts
+++ b/noodles-editor/src/noodles/__migrations__/008-rename-date-to-datetime.test.ts
@@ -1,0 +1,247 @@
+import { describe, expect, it } from 'vitest'
+import type { NoodlesProjectJSON } from '../utils/serialization'
+import { down, up } from './008-rename-date-to-datetime'
+
+const createProjectWithDateOp = (): NoodlesProjectJSON => ({
+  version: 7,
+  nodes: [
+    {
+      id: '/date-1',
+      type: 'DateOp',
+      position: { x: 100, y: 100 },
+      data: {
+        inputs: {
+          date: '2024-01-01T00:00:00',
+        },
+      },
+    },
+  ],
+  edges: [],
+  viewport: { x: 0, y: 0, zoom: 1 },
+  timeline: {},
+})
+
+const createProjectWithDateTimeOp = (): NoodlesProjectJSON => ({
+  version: 8,
+  nodes: [
+    {
+      id: '/date-1',
+      type: 'DateTimeOp',
+      position: { x: 100, y: 100 },
+      data: {
+        inputs: {
+          date: '2024-01-01T00:00:00',
+        },
+      },
+    },
+  ],
+  edges: [],
+  viewport: { x: 0, y: 0, zoom: 1 },
+  timeline: {},
+})
+
+const createProjectWithMultipleOps = (): NoodlesProjectJSON => ({
+  version: 7,
+  nodes: [
+    {
+      id: '/date-1',
+      type: 'DateOp',
+      position: { x: 100, y: 100 },
+      data: {
+        inputs: {
+          date: '2024-01-01T00:00:00',
+        },
+      },
+    },
+    {
+      id: '/number-1',
+      type: 'NumberOp',
+      position: { x: 200, y: 200 },
+      data: {
+        inputs: {
+          value: 42,
+        },
+      },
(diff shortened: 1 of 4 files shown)


akre54:
Could we make this a different field / fieldComponent / op? This feels like it's overloading the operator a bunch...

chrisgervang:
I don't see this as an overload - it's just adding precision, which is natively supported by the built-in UI and Temporal.. what's overloaded?

Prior to this change it looks like this: 
Screenshot 2025-11-14 at 3 20 22 PM


It's common for someone using date times to need this level of precision - minutes aren't precise enough.

chrisgervang:
What additional set of ops are you imagining?

We haven't incorporated timezone.. we could make an op for any number of the 4 Temporal types (Instant, PlainDate, PlainDateTime, ZonedDateTime)

akre54:
Yes those were along the lines of what I was thinking. There start to be *a lot* of combinatorics around potential extra features you might want to add (timezones, dates vs datetime, and sub-seconds specifically), while we'd want to allow downstream operators to work with them all.

I hear you that it's not a big deal to add *one small* feature, but it's a really slippery slope and makes the Date op less useful for the folks who *don't* need micro-seconds. The way that forcing everything to use DateTime, or ZonedDateTime isn't useful to those who just need a date.

I'd rather have separate ops / fields for these various types, and then either a Generic date data field type, a Superset field type, or add some sort of Union Field support for these cases

chrisgervang:
Yeah I think that's the right design too. I do think that if you're using a DateTime, then the time precision is necessary. Sounds like the "Date" op should be renamed to "DateTime" and then we can add new ops for each use case. I think that's better done in a follow up - for now, this adds the full date-time support I'd expect when working with ISO8601-like timestamps

noodles-editor/src/noodles/fields.ts
```diff
@@ -522,7 +522,15 @@ export class BooleanField extends Field<z.ZodBoolean> {
   }
 }
 
-export class DateField extends Field<z.ZodUnion<[z.ZodDate, z.ZodISODateTime]>> {
+export class DateField extends Field<
```
akre54: Is this now a catch-all?

chrisgervang: As written, I meant to rename it to `DateTime` since the field outputs a `PlainDateTime`.

It's easy to make a catch-all for the inputs. It outputs only one type and can have a pure transform.

Seems like we could either have a field for each (TimeField, DateField, DateTimeField, and ZonedDateTimeField), or maybe we could have a field that outputs the full type, a ZonedDateTime, and uses a transform to convert it to a subset type. We'd need to make sure the up and down conversation pass through with tests

chrisgervang: The four value types could serialize differently too. A combined field type would need to coerce the value into a zoned date time string in the project file.

I'm wondering how the filed component would reconfigure itself for each picker? The Point2D or Color field's transform are the same input for either transform, but that's not the case for the 4 types of date/times

It might be cleaner to just have 4 field types, and implement the fields to accept any input.

Is there an important reason to have one field instead, or are the types fundamentally different and so they should each have their own field?

chrisgervang:
Going to merge this for now to unblock project that need the precision, but we've also discussed ideas for specs to move forward with. 

Our ideal would be to have a single field (i.e. a TemporalField, currently called DateField) that can take any input, and be able to customize it's output schema for each op that uses it.

We could, say, the have an option for changing it's temporal type and precision in UI.

```
"[type]-[precision]"
```

Where types are:

- date: Temporal.PlainDate
- time: Temporal.Instant
- datetime: Temporal.PlainDateTime
- zoned-datetime: Temporal.ZonedDateTime

And precision defines the level of granularity offered in the UI:
- y: year
- m: month
- d: day
- h: hour
- s: second
- ms: millisecond

For example, `new TemporalField({ precision: 'date-d' })` would map do Temporal.PlainDate with day precision (year, month, day).

Only combinations that make sense for a given type would be allowed (i.e. `date-d` is allowed but `date-s` isn't since the underlying type does not support seconds).

It's a lot of options, however, I think this does a good job at keeping the number of classes down and avoids risky conversions or needing to make assumptions about how to store data (i.e. if precision was `time-s` but mapped to PlainDateTime you'd need to store an arbitrary date value (jan 1, 1970?) when all you wanted was to store a time value)

PR #177, opened 2025-12-15
feat: Add react-color ChromePicker to color field component

chrisgervang:
## Summary

Replaces the basic HTML color input with the react-color ChromePicker, adding:
- **Copy/paste support** for hex color values
- **Alpha channel support**

### Field Value Format

ColorField now **always** stores values as **8-character hex strings** (e.g., `#00ff00ff` including alpha channel). The schema includes a transform that automatically normalizes 6-character hex strings to 8-character format, ensuring old project files are automatically upgraded when loaded.

### Array Color Handling (Deck.gl Integration)

When ColorField is created with `transform: hexToColor`, the field stores **array values** `[r, g, b, a]` instead of hex strings. This is used extensively for Deck.gl layer color properties:

```typescript
// Example from PathLayerOp
getColor: new ColorField('#006ac6', { accessor: true, transform: hexToColor })
```

**How it works:**

1. **Input**: User provides hex string `#006ac6` (or 8-char `#006ac6ff`)
2. **Schema Transform**: ColorField's schema normalizes to 8-char hex
3. **Field Transform**: The `transform: hexToColor` option applies after schema validation, converting hex to array
4. **Internal Value**: Field stores `[0, 106, 198, 255]` as an array
5. **Serialization**: When saving to project files, `serialize()` converts back to hex string `#006ac6ff`

**Why arrays?** Deck.gl layer properties expect color arrays `[r, g, b, a]` where each component is 0-255. The transform option allows the field to store the value in Deck.gl's native format while serializing as human-readable hex strings in project files.

### Serialization Flow

```
User Input → Zod Schema (normalize to 8-char hex) → Field Options Transform (optional: hex → array) → Internal Storage
                                                                                                              ↓
Project File ← serialize() (convert array → hex if needed) ← Internal Storage
```

**What's stored in project files?**
- Always 8-character hex strings (e.g., `#00ff00ff`)
- Never arrays (arrays are converted to hex in `serialize()`)
- This makes project files human-readable and easy to diff

### Backwards Compatibility

- **Old project files** with 6-char hex (e.g., `#00ff00`) are automatically upgraded to 8-char when loaded
- **No migration needed** - the schema transform handles it transparently
- **Array handling** in serialize() is still necessary because fields with `transform: hexToColor` store arrays internally and need to convert them to hex for project files

## Testing

- Added comprehensive unit tests for hex string serialization
- Added tests for array color handling with `transform: hexToColor`
- Verified backwards compatibility with 6-char hex values

## Breaking Changes

None. The change is backwards compatible:
- Old project files with 6-char hex are automatically upgraded
- The UI change enhances the user experience while maintaining all existing functionality


chrisgervang:
> Nice. Can you add a screenshot?

Still iterating, but atm I'm just using the picker that looks like chrome's built-in picker since it accepts RGB, HEX, HSB, etc.. the problem with the real built-in picker is it doesn't support copy/paste. react-color does support it, and supports adding color presets if we want them later on.

akre54:
Can you open a patch against the library to allow the hex input when alpha isn't 1? I think you said you know the guy who wrote it

noodles-editor/src/noodles/components/field-components.tsx
```diff
@@ -1153,26 +1162,169 @@ export function ColorFieldComponent({
...
+    let top = swatchRect.bottom + margin
+    let left = swatchRect.left
+
+    // Vertical positioning: prefer below, but use above if not enough space
+    if (spaceBelow < pickerHeight + margin && spaceAbove > spaceBelow) {
+      top = swatchRect.top - pickerHeight - margin
+    }
+
+    // Horizontal positioning: ensure picker stays within viewport
+    if (left + pickerWidth > viewportWidth) {
```
akre54: This is a lot of custom logic

chrisgervang: It's complicated a bit since the swatch scales size with the node but the picker doesn't - you want the picker to render large no matter the viewport. 

Image

Do you know a better way to do this?

chrisgervang:
> The hex input change seems to be a conscious decision, so it might be nice to feed back some of these changes

He's open to it! The library always represents color with alpha internally and so this would just be a small UI change.

> I do think we need to come up with an overarching philosophy about type precision and metadata. We have a bunch of one offs and edge cases that can compound in interesting ways. I think there isn't a simple answer and there's a reason programming languages disambiguate these concepts.
>
> We have some leeway to trade-off execution speed and configurability for ease-of-use, but it'd be nice to be consistent about this

I hear you wrt these precision/model situations. Many recent changes touch on it and I'm curious what choices we have. JS tends to pick an internal type with the broadest support and then downsamples - `Number` is _always_ a double, but can be rounded to an int-like. Color in webgl, canvas 2d, and css is always 4 channel, but can accept a 3 channel input. The `Date` class attempted to do this too.. I find it interesting that `Temporal` is designed 5 core classes that do not convert cleanly to each other. That choice seems intentional. And surely, there are trade-offs to picking the broadest type (memory, frequent conversion, etc)

noodles-editor/src/noodles/components/field-components.tsx
chrisgervang: I think this is the most robust way to do it - it's using a portal and position fixed, and then just getting the position of the swatch to position the picker relative to it, and accounts for the windows edges.

The graph needs the vertical guard and I could see the property panel using the horizontal guard as we move more away from theatre.

chrisgervang:

Being able to fade individual elements in a layer by supplying alpha to each `getColor` datum is an important use case for fading datum in and out

PR #181, opened 2025-12-15
fix: Preserve operator names in production builds

chrisgervang:
## Summary

Fixes minified operator names in production builds where class names get mangled by the minifier, causing the UI to display incorrect operator names, categories, and colors.

**Root Cause:** Code was relying on `constructor.name` which gets minified (e.g., `ScatterplotLayerOp` → `Zt`).

**Solution:** Use static `displayName` and `description` properties accessed via `(constructor as typeof Operator).property` pattern. Static properties with string literals survive minification.

### Core Architecture

**Static Properties for Metadata:**
```typescript
export class NumberOp extends Operator<NumberOp> {
  static displayName = 'Number'
  static description = 'A numeric constant'
  // ...
}
```

**Access Pattern Throughout Codebase:**
```typescript
// Extract from constructor for minification-safe access
const { displayName, description } = operator.constructor as typeof Operator
const { type } = field.constructor as typeof Field
```

### Category System

Categories now use display names (without "Op" suffix):
```typescript
export const categories = {
  data: ['File', 'Filter', 'Merge', ...],
  layer: ['ScatterplotLayer', 'PathLayer', ...],
  // ...
}
```

Helper function converts between ReactFlow's suffixed types and display names:
```typescript
export function nodeTypeToDisplayName(nodeType: string): string {
  return nodeType.replace(/Op$/, '')
}
```

**Example:** `node.type` is `"ScatterplotLayerOp"` → converts to `"ScatterplotLayer"` → matches category entry.

### Special Case: ForLoop

"ForLoop" is a synthetic menu type that creates two actual nodes:
- Menu shows: `'ForLoop'`
- Creates nodes: `'ForLoopBeginOp'` and `'ForLoopEndOp'`
- Categories include: `['ForLoop', 'ForLoopBegin', 'ForLoopEnd']`

This works because `nodeTypeToDisplayName('ForLoopBeginOp')` → `'ForLoopBegin'` which matches the category.

## Files Changed

**Core:**
- `operators.ts` - Import Operator type for constructor casting
- `fields.ts` - Remove instance getter for type property
- `categories.ts` - Convert all entries to display names (without "Op")

**Components:**
- `op-components.tsx` - Convert node types to display names for lookups
- `node-properties.tsx` - Extract static properties from constructor
- `node-tree-sidebar.tsx` - Use static displayName access  
- `field-components.tsx` - Extract field type from constructor

**Hooks & Utils:**
- `use-project-modifications.ts` - Extract field types for error messages
- `mcp-tools.ts` - Simplified to use operator directly

**Tests:**
- Updated to use static access: `NumberOp.displayName` instead of `op.displayName`

### Automated
- ✅ All 182 tests passing
- ✅ Production build succeeds
- ✅ TypeScript compilation clean

### Manual Testing

**Production Build:**
```bash
cd noodles-editor
yarn build
yarn preview
```

**Verify:**
1. **Block Library** - Categories show correct colors and names
2. **Node Headers** - Each node displays correct operator name with category color
3. **Property Panel** - Selected nodes show correct type and description
4. **Node Tree Sidebar** - All operators display with correct names
5. **Existing Projects** - Load examples and verify operators work correctly

## Why This Approach

**Advantages:**
- ✅ **Minification-safe**: Static properties survive build process
- ✅ **Type-safe**: TypeScript enforces correct property access
- ✅ **Clean**: Pattern is consistent throughout codebase
- ✅ **Flexible**: Can access both statically (`NumberOp.displayName`) and from instances

**Trade-offs:**
- Requires `(constructor as typeof Operator)` cast for instance access
- More verbose than instance property, but clearer intent

## Follow Up

- Migrate ReactFlow and serialized projects to use display names instead of "Op" suffixed types

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -557,7 +557,7 @@ Any field can be keyframed via Theatre.js. Changes in timeline propagate through
 3. Define outputs with `createOutputs()` method
 4. Implement `execute()` method with pure function logic
 5. Register operator in operator registry
-6. Add to category in `components/categories.ts`
+6. Add to category in `components/categories.ts` using display name without "Op" suffix (e.g., `'File'` not `'FileOp'`)
 7. **Write unit tests** (required for all operators)
 8. Document behavior and limitations if complex
 9. Test in UI with example projects
--- a/dev-docs/developing.md
+++ b/dev-docs/developing.md
@@ -105,7 +105,7 @@ The project uses:
 - Write unit tests for operators
 - Use TypeScript strictly
 - Comment complex logic
-- **Never use `constructor.name`** - it gets minified in production builds. Use static `displayName`/`description` properties and access them via `(operator.constructor as typeof Operator).displayName`
+- **Avoid `constructor.name`** - it gets minified in production builds. Use static `displayName`/`description` properties and access them via `(operator.constructor as typeof Operator).displayName`
 
 ### Maintenance
 


noodles-editor/src/noodles/components/node-tree-sidebar.tsx
```diff
@@ -27,10 +27,16 @@ interface TreeNode {
...
 // Build hierarchical tree from flat operator list
-function buildTree(operators: Map<string, Operator<IOperator>>): TreeNode[] {
+function buildTree(
+  operators: Map<string, Operator<IOperator>>,
+  reactFlowNodes: Array<{ id: string; type: string }>
+): TreeNode[] {
   const tree: TreeNode[] = []
   const nodeMap = new Map<string, TreeNode>()
 
+  // Create a map of id -> type from ReactFlow nodes for fast lookup
```
chrisgervang: Before:
Image

After:
Image

akre54:
Why is this necessary? Can't we just use the `displayName` that exists already, or do a mapping of one to the other (option 2). I think the cons aren't true. Also fine with option 3

Also - I don't know that we need the op category in the left tree. It seems like a lot of extra noise.

noodles-editor/src/noodles/components/node-properties.tsx
```diff
@@ -202,7 +202,7 @@ function NodeProperties({ node }: { node: NodeJSON<unknown> }) {
       <div className={s.header}>
         <div className={s.title}>
           {displayName}
-          <div className={cx(s.capsule, headerClass(name))}>{typeCategory(name)}</div>
+          <div className={cx(s.capsule, headerClass(type))}>{typeCategory(type)}</div>
```
akre54: So why does this work in production currently?

chrisgervang: The only parts that work in production are using `const opTypes = {}` map which is stringified before minification. Nothing in production referencing `constructor.name` are working - they're just subtle issues

chrisgervang: Image

The badges are all "Unknown"

akre54: Ah. And what are the node headers using / why does that work?

akre54: Looked into it and apparently they're using `name` from the constructor

chrisgervang:
> Static is definitely type safe

That's not always true. It's only type safe when you do `ArcOp.displayName` not `op.constructor.displayName` - the types fall apart since typescript types this loosely as a `Function`.

Static has real limitations, so while its the textbook-correct OOP feature to use, we don't want it here. There are more benefits to using an instance property because of our access patterns. `readonly` is only a typescript feature, but its what is typically used in this case

noodles-editor/src/noodles/components/op-components.tsx
```diff
@@ -148,7 +148,10 @@ export function typeDisplayName(type: NodeType) {
 export function getNodeDescription(type: NodeType): string {
   // Check for regular operators first
   if (type in opTypes) {
-    return opTypes[type]?.description || ''
+    // Create a temporary instance to access instance property
+    const ctor = opTypes[type]
```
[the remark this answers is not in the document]
chrisgervang: Sure, I don't want to merge this. But that's only true if its a static or in a `get()`. Description is now defined as a readonly property. This is the only place we use it off the class rather than the instance. The `op.constructor` pattern isn't correct either as its not type safe - what do you suggest we do?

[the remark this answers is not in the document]
chrisgervang: Here's an option that seems like a good balance:

```
abstract class Op {
  static readonly description: string = ''

  get description(): string {
    // typed access to static via the runtime ctor
    return (this.constructor as typeof Op).description
  }
}

class ArcOp extends Op {
  static readonly description = 'Draws an arc'
}

const ctor = opTypes[type]
return ctor.description ?? ''
```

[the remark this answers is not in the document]
akre54: > Here's an option that seems like a good balance:

Yes I made that suggestion below. I don't think it's a necessary change and I'd rather we not merge it but I don't feel particularly strongly.

Adding `readonly` I'm fine with but also don't particularly care for similar reasons

chrisgervang:
> That just isn't correct. You can cast (op.constructor as typeof Operator).displayName. It's just not something we've cared about.

> If we really wanted to make this cleaner we could add getters on the Operator base class for displayName / description to desugar as return (this.constructor as typeof Operator).displayName if we're using this a whole lot.

Casting is an escape hatch - it's not correct to say op.constructor is type safe. But regardless, I think we're converging on the same idea - add getters that cast to the base class.

noodles-editor/src/noodles/components/op-components.tsx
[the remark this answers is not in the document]
chrisgervang: What did you mean by "You can just use the `prototype` to access instance properties"? I interpreted that as meaning you suggesting this is valid:

```
class ArcOp  {
  static description = 'Draws an arc'
}

const op = new ArcOp()

op.prototype.description // 'Draws an arc'
ArcOp.prototype.description // 'Draws an arc'
```

But trying these in chrome dev tools doesn't work.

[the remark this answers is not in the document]
chrisgervang: > Yes I made that suggestion below. I don't think it's a necessary change and I'd rather we not merge it but I don't feel particularly strongly.

Let's just do this and move on - the root cause of this issue had nothing to do with the static values and so I was just looking to improve their types while I'm in here. The lesson is `op.constuctor.name` is incorrect to use because it gets minified.

chrisgervang:
> I don't think it's an escape hatch, it's the way typescript works. It's typesafe because op.constructor is an Operator class but it's also most correct to say that it's a js Function. If you tried to cast to op.constructor to any other class it would fail. It's just untyped by default due to a design decision to align closer to JS and due to the way casting works. That's what I mean by type safety

Gotcha - I didn't know that, thanks for explaining it.

I'll stick to casting at each usage then. We're gonna want to use `typeof Operator` to give it the class type rather than the instance type. Now they're all consistent, static, and typed - and working in production. As a bonus, I was able to fix some other broken types too.

@greptile please re-review

noodles-editor/src/noodles/components/categories.ts
```diff
@@ -1,118 +1,124 @@
...
+    'SplitMapViewState',
   ],
-  widget: ['FpsWidgetOp'],
+  widget: ['FpsWidget'],
 } as const
+
+// TODO: Remove this function when we fully migrate to operator displayNames
+export function nodeTypeToDisplayName(nodeType: string): string {
+  return nodeType.replace(/Op$/, '')
+}
```
chrisgervang: Follow up: #181 switched categories to use `displayName`. Do we also want to migrate ReactFlow and serialized project files to use display names instead of "Op" suffixed types?

Context is `opTypes` keys stringify at build-time to their class name

https://github.com/joby-aviation/noodles.gl/blob/86807a5b813bb483edde5463f8ff99804c22b7ef/noodles-editor/src/noodles/operators.ts#L5303-L5309

And it was assumed `op.constructor.name` would be equivalent in production, but those values get minified. `displayName` already existed and by convention is a unique name we could use to identify an op.

PR #182, opened 2025-12-15
feat: Implement TemporalField with configurable types and precision

### LINKED ISSUE #146: Refactor DateField for precision and temporal type control
Going to merge this for now to unblock project that need the precision, but we've also discussed ideas for specs to move forward with. 

Our ideal would be to have a single field (i.e. a TemporalField, currently called DateField) that can take any input, and be able to customize it's output schema for each op that uses it.

We could, say, the have an option for changing it's temporal type and precision in UI.

```
"[type]-[precision]"
```

Where types are:

- date: Temporal.PlainDate
- time: Temporal.Instant
- datetime: Temporal.PlainDateTime
- zoned-datetime: Temporal.ZonedDateTime

And precision defines the level of granularity offered in the UI:
- y: year
- m: month
- d: day
- h: hour
- s: second
- ms: millisecond

For example, `new TemporalField({ precision: 'date-d' })` would map do Temporal.PlainDate with day precision (year, month, day).

Only combinations that make sense for a given type would be allowed (i.e. `date-d` is allowed but `date-s` isn't since the underlying type does not support seconds).

It's a lot of options, however, I think this does a good job at keeping the number of classes down and avoids risky conversions or needing to make assumptions about how to store data (i.e. if precision was `time-s` but mapped to PlainDateTime you'd need to store an arbitrary date value (jan 1, 1970?) when all you wanted was to store a time value)

There's a lot of room to iterate on this spec too, maybe we support fewer types or precisions - or go a different direction all together. Temporal types are tricky, but I think we're positioned to give each variation a nice UI and I expect zod to do a lot of the heavy lifting for us.

_Originally posted by @chrisgervang in https://github.com/joby-aviation/noodles.gl/issues/142#issuecomment-3549589010_

chrisgervang:
## Summary

Implements the TemporalField as specified in #146 to provide flexible and precise temporal type handling.

Closes #146

## Features

**Multi-Type Support:**
- `date` - Temporal.PlainDate
- `time` - Temporal.PlainTime
- `datetime` - Temporal.PlainDateTime
- `zoned-datetime` - Temporal.ZonedDateTime

**Configurable Precision:**
- Year (`y`), Month (`M`), Day (`d`)
- Hour (`h`), Minute (`m`), Second (`s`), Millisecond (`ms`)

**Easy Configuration:**
```typescript
new TemporalField("datetime-s")  // Format string
new TemporalField({ type: "date", precision: "d" })  // Object config
```

**Validation:**
- Type-precision combinations validated
- Date fields: precision up to day only
- Time fields: precision from hour onwards
- Datetime/zoned-datetime: all precisions supported

## Integration

✅ Zod schemas for type-safe validation  
✅ UI component with HTML5 input types  
✅ Theatre.js bindings for timeline animation  
✅ Serialization/deserialization support  
✅ Comprehensive test coverage (115 tests)

## Files Modified

- `noodles-editor/src/noodles/fields.ts`
- `noodles-editor/src/noodles/fields.test.ts`
- `noodles-editor/src/noodles/operators.ts`
- `noodles-editor/src/noodles/components/field-components.tsx`
- `noodles-editor/src/noodles/theatre-bindings.ts`

## Testing

- [ ] Run tests: `yarn test`
- [ ] Create operator with TemporalField
- [ ] Test different type-precision combinations
- [ ] Verify timeline animation works
- [ ] Test serialization/deserialization

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>


noodles-editor/src/noodles/components/field-components.tsx
```diff
@@ -1134,6 +1136,185 @@ export function DateFieldComponent({
...
+  // Determine HTML input type based on temporal type
+  const getInputType = (): string => {
+    switch (temporalType) {
+      case 'date':
+        return 'date'
+      case 'time':
+        return 'time'
+      case 'datetime':
+      case 'zoned-datetime':
+        return 'datetime-local'
```
akre54: I'm not sure this works. We'd need inputs to be two-part: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/datetime-local#setting_timezones

noodles-editor/src/noodles/operators.ts
```diff
@@ -694,9 +695,9 @@ export class BooleanOp extends Operator<BooleanOp> {
   }
 }
 
-export class DateTimeOp extends Operator<DateTimeOp> {
-  static displayName = 'DateTime'
-  static description = 'A date and time'
+export class LegacyDateTimeOp extends Operator<LegacyDateTimeOp> {
```
chrisgervang: Keeping around temporarily for testing

akre54:
I'm a little concerned about all the up-and-downsampling.

More generally would it be better if we used unixtime ([`Temporal.PlainDateTime`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/PlainDateTime) / [`Temporal.Instant`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Instant)) as the base type and only used the precision options to choose what UI to render? Feels like an extension of the philosophy we discussed on #192

noodles-editor/src/noodles/operators.ts
```diff
@@ -1127,8 +1218,8 @@ export class CategoricalColorRampOp extends Operator<CategoricalColorRampOp> {
   }
 }
 
-export class TimeOp extends Operator<TimeOp> {
-  static displayName = 'Time'
+export class AnimationTimeOp extends Operator<AnimationTimeOp> {
```
akre54: Why change this?

chrisgervang:
> I'm a little concerned about all the up-and-downsampling.
> 
> 
> 
> More generally would it be better if we used unixtime ([`Temporal.PlainDateTime`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/PlainDateTime) / [`Temporal.Instant`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Instant)) as the base type and only used the precision options to choose what UI to render? Feels like an extension of the philosophy we discussed on #192

Is the question whether or not we have an op for each Temporal class, or if every temporal value is mapped to base type?

I'm curious too. If we map our fields to Temporal each type then we're able to simplify our implementation to simple wrappers around built-in classes. If we didn't (say, normalized everything to an Instant or ZonesDateTime) then we'd need to form our own opinions on how format the output and convert.

For example with "time" (no date) mapped to an Instant.. would we assume a date of Jan 1 1970? Inverse, for Date would we assume a time of midnight? Instant requires both whereas PlainDate or Time classes do not.

Another example is DirectionsOp - [Mapbox](https://docs.mapbox.com/api/navigation/directions/) takes certain inputs for driving traffic - PlainDateTime, Instant ("Z"), or Instant with an hour offset. The PlainDateTime variant scales well across projects, since their backend localizes using the origin's location. But [Google Maps transit](https://developers.google.com/maps/documentation/routes/transit-route) only accepts Instant - in practice, that means the field should be an Instant type or ZonedDateTime and converted to UTC, and I'd prefer to use a ZonedDateTime op to make sure I'm requesting the correct local time (5pm in SF vs 5pm in NYC)

My current leaning is move precision into a field to dynamically configure the input UI and rounding, and keep the "type" variants mapped to Temporal classes. 

This puts a category of up/down sampling on the op level, keeps the grammar closer to JS, and would reject ambiguous field connections (I.e. Time to Date)

Another alternative is to not support precisions that cut out date or time (in other words, we only support date+time). We then just need utilities to convert between Plain and Instant or Zoned. Applications would need to choose their own assumption for date or time if they just need one of these components (I'd prefer that over silently assuming Jan 1 1970 for them)

I do think these are the right questions to be asking - do we really need to support all type and precision variations, and where is it ok to decide "even though it's technically available, we're deciding we shouldn't support it in the grammar"?

akre54:
What's the use case for having multiple levels of precision? Just for formatting? Or date math?

chrisgervang:
I only have a use case for having the maximum level of precision. I remember you were advocating support for the option to have less precision which motivated the spec proposed here, is that no longer a use case for you?

akre54:
It's very possible but I don't recall that discussion. Mind slacking me a link if you find it? I'm curious

akre54:
Ah, probably [this comment](https://github.com/joby-aviation/noodles.gl/pull/142#issuecomment-3535001330). I imagine that comment was mostly about making the UI simpler. Agreed the data should be stored with higher precision. Maybe it's just about making the UI configurable?

chrisgervang:
Serializing the UI state seems to be adding complexity, if it's stored on the node. Could be simplifier to store a global UI config

Another exposition in this PR is reviewing the various Temporal types and how these might be expressed as ops. I don't think we typically need _all_ of the Temporal types, but do need to define assumptions when converting between them.

chrisgervang:
I personally wouldn't use the configurable precision feature, but would use multi-type support because data comes in to projects in all types, and outputs may need to be in any type

akre54:
We tend to take the tack that types propagate through the graph unless there is specifically a reason to convert them to a different type(say like a BoundsOp) or the operator itself needs the transformation (Deck LayerOps for instance).

I'm imagining this would work similar to boxing and unboxing in a type systhem, and I'd assume it would work the same here. But for sure we should have it well documented and tested

PR #198, opened 2026-01-05
fix: Theatre.js spurious keyframes on selection and connected fields

chrisgervang:
## Problem

Theatre.js was creating unwanted keyframes in two scenarios:
1. **When selecting nodes**: Field subscriptions would fire and attempt to sync unchanged values to Theatre
2. **When fields were connected to animated upstream operators**: Every frame update would try to create keyframes

This happened because:
- Object types (Color, Vec2, Vec3, Point2D, Point3D) use reference comparison which always returns false for new objects
- Theatre.js adds function properties (like toString) to RGBA objects, causing equality checks to fail
- Connected fields were being synced even though their values come from upstream operators

## Solution

1. **Deep equality check**: Added isObjectEqual() helper that:
   - Filters out function properties before comparing
   - Performs proper deep equality check for object properties

2. **Normalize vector/point formats**: Convert arrays to objects before comparison:
   - [0, 0] → {x: 0, y: 0} for Vec2
   - [lng, lat] → {lng, lat} for Point2D
   - etc.

3. **Skip connected fields**: Check if field has incoming edge before syncing to Theatre
   - If a field is driven by an upstream operator, don't sync its reactive updates

4. **Edge store**: Created edge store to make React Flow edges accessible to theatre-bindings

## Testing

1. Select different nodes with keyframed color fields - no spurious keyframes should be created
2. Connect animated fields to downstream operators - downstream fields shouldn't create keyframes
3. Manually keyframe colors/vectors - should still work normally

## Files Changed

- theatre-bindings.ts: Add equality checks, normalization, connected field detection
- store.tsx: Add edge store
- noodles.tsx: Sync edges to store

## TODO:

- [ ] add tests
- [ ] future PR: replace react-flow edge store with ours, which can be accessed outside of a react context


chrisgervang:
These changes improve performance as well when scrubbing the timeline

PR #201, opened 2026-01-05
fix: GlobeView text rendering with cullMode parameter

chrisgervang:
## Problem

Text rendering in `GlobeView` is not visible from any angle due to triangle culling (https://github.com/visgl/deck.gl/issues/9777). The default culling mode for GlobeView (`'back'`) causes text to be culled and invisible.

## Solution

Add `parameters.cullMode` to TextLayerOp and GeoJsonLayerOp with options:
- `'back'`: Back-face culling
- `'front'`: Front-face culling
- `'none'`: Disable culling - fixes GlobeView text rendering, and is the default for none-GlobeView.
  - See https://luma.gl/docs/api-reference/core/parameters#assembly-culling

Long term, we'll need to solve these issues in deck but the root cause needs to be found, and solution approaches discussed.

## Changes

- Added `parameters` CompoundPropsField to TextLayerOp
- Added `parameters` CompoundPropsField to GeoJsonLayerOp
- Each contains `cullMode` StringLiteralField with three options

## Testing

1. Create a TextLayer or GeoJsonLayer with text in GlobeView
2. Observe that text is not visible with default settings
3. Set `parameters.cullMode` to `'none'` to make text visible
4. Verify default (`'none'`) behavior is preserved for non-GlobeView use cases

## References

- https://github.com/visgl/deck.gl/issues/9777

### DIFF
--- a/noodles-editor/src/noodles/operators.ts
+++ b/noodles-editor/src/noodles/operators.ts
@@ -3305,7 +3305,7 @@ export class TextLayerOp extends Operator<TextLayerOp> {
       }),
       extensions: new ListField(new ExtensionField()),
       parameters: new CompoundPropsField({
-        cullMode: new StringLiteralField('back', {
+        cullMode: new StringLiteralField('none', {
           values: ['none', 'back', 'front'],
         }),
       }),
@@ -3612,7 +3612,7 @@ export class GeoJsonLayerOp extends Operator<GeoJsonLayerOp> {
       _full3d: new BooleanField(false),
       extensions: new ListField(new ExtensionField()),
       parameters: new CompoundPropsField({
-        cullMode: new StringLiteralField('back', {
+        cullMode: new StringLiteralField('none', {
           values: ['none', 'back', 'front'],
         }),
       }),


noodles-editor/src/noodles/operators.ts
```diff
@@ -3606,6 +3611,11 @@ export class GeoJsonLayerOp extends Operator<GeoJsonLayerOp> {
       elevationScale: new NumberField(1, { min: 0, max: 100 }),
       _full3d: new BooleanField(false),
       extensions: new ListField(new ExtensionField()),
+      parameters: new CompoundPropsField({
+        cullMode: new StringLiteralField('back', {
```
chrisgervang: TODO: default should be `'none'`

And, we could fix this by hard-coding `'none'` rather than adding a parameter - and then revisit after teh upstream issue is fixed.

Related #35

charlieforward9:
Any blocker to merge? Running into this.

chrisgervang:
> Any blocker to merge? Running into this.

No blockers. The upstream fix is none-trivial so this param will need to be exposed for now. Are you using globe view?

charlieforward9:
Thanks for linking the related issue. Yes I am using GlobeView.

PR #203, opened 2026-01-05
Add MergeOpsOp operator

chrisgervang:
## Summary

Adds MergeOpsOp operator that collects values from connected operators into an object with operator names as keys.

[merge-ops.zip](https://github.com/user-attachments/files/24426976/merge-ops.zip)

## Motivation

When working with multiple related operators (e.g., different data sources, multiple computed values), it's useful to collect them into a single object structure. This operator automatically uses operator names as keys, making the resulting object more semantic than array indices.

## Changes

- Added `MergeOpsOp` operator class with ListField input
- Uses edge store to auto-detect connected operator IDs
- Extracts base names from operator paths (e.g., `/data/loader` → `loader`)
- Supports both static values and accessor functions
- Added to data category in operator menu
- Added necessary imports (`getEdgeStore`, `getBaseName`)

## Usage Example

Connect multiple operators to the `values` list input. MergeOpsOp will output an object like:

```javascript
{
  "temperature": 72,
  "humidity": 65,
  "pressure": 1013
}
```

Where keys come from the connected operator names.

## Dependencies

This PR is based on #198 (Theatre.js fix) which adds the edge store functionality.

## Testing

1. Create several operators (e.g., NumberOp instances named "temperature", "humidity", "pressure")
2. Create a MergeOpsOp
3. Connect the operators to MergeOpsOp's values list
4. Verify output object uses operator names as keys
5. Test with accessor functions to ensure accessor mode works correctly

### DIFF
--- a/noodles-editor/src/noodles/components/categories.ts
+++ b/noodles-editor/src/noodles/components/categories.ts
@@ -31,6 +31,7 @@ export const categories = {
     'Filter',
     'Geocoder',
     'Merge',
+    'MergeOps',
     'Network',
     'Out',
     'RandomizeAttribute',
--- a/noodles-editor/src/noodles/operators.ts
+++ b/noodles-editor/src/noodles/operators.ts
@@ -179,12 +179,12 @@ import {
   WidgetField,
 } from './fields'
 import { DEFAULT_LATITUDE, DEFAULT_LONGITUDE, safeMode } from './globals'
-import { getAllOps, getOp, hasOp } from './store'
+import { getAllOps, getEdgeStore, getOp, hasOp } from './store'
 import { composeAccessor, isAccessor } from './utils/accessor-helpers'
 import type { ExtractProps } from './utils/extract-props'
 import { projectScheme } from './utils/filesystem'
 import type { OpId } from './utils/id-utils'
-import { isDirectChild } from './utils/path-utils'
+import { getBaseName, isDirectChild } from './utils/path-utils'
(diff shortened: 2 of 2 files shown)


akre54 (changes requested):
This feels pretty brittle and "magical" to me since it breaks a number of assumptions in the rest of the app, and now makes ids semantically meaningful.

I understand the motivation but let's maybe try to find prior art for how other VPLs handle this type of design issue

chrisgervang:
Agreed! I'm open to exploring other ideas - demonstrating the motivation and potential is my main goal at this point. I was curious to see how this would work in-practice, and I just needed _something_ to work for doing the titles in the 2025 flights.

A big drawback of this solution is I couldn't use multiple mergeops on the same level since the input id's must be unique (e.g. string, string-1, string-2) - I had to embed each "datum" into their own container (e.g `datum-1/string`, `datum-2/string`). This meant I had to do a lot of "up and into" navigating to adjust values.

And another is my sidebar of nodes got very long.

What I really want is a "bag of fields" op that can be customized, like other VPLs have. With the ability to add/remove fields. And then I want to simply copy/paste this op for each "datum" - we've talked about this before, would be a good time to add it!

Until then, my workaround has been to make a CodeOp with references to ops placed next to it - but code reference subscription within container are currently not subscribing on-load (mentioned in my dev log).. that'd be a great issue to fix since containers have quickly become a go-to I use in every project.

PR #204, opened 2026-01-05
Add multiline textarea support to StringField

chrisgervang:
## Summary

Adds multiline textarea support to StringField and makes StringOp resizable for better multi-line string editing.

Before:

After:

Screenshot 2026-01-04 at 9 05 23 PM

### StringField (fields.ts)
- Added `multiline` option to constructor
- Automatically uses textarea for fields with multiline=true or values containing newlines

### Field Rendering (field-components.tsx)
- Render textarea instead of input for multiline strings
- Track node height to dynamically resize textarea
- Auto-size textarea to fill node (nodeHeight - 40px for header)

### StringOp Node (op-components.tsx)
- Added to `resizeableNodes` array
- Minimum height: 100px
- Allows users to resize node for comfortable multiline editing

### Styles (noodles.module.css)
- Added `fieldTextarea` styles
- No scrollbar, auto-resizing height
- Consistent with input field styling

### StringOp Default (operators.ts)
- Enabled `multiline: true` by default for StringOp

## Benefits

- Better UX for editing long strings or multi-line content
- Resizable node adapts to content needs
- Automatic detection of newlines triggers textarea mode
- Clean, native textarea experience

## Testing

1. Create a StringOp node
2. Resize the node vertically
3. Enter multi-line text and verify textarea expands
4. Verify existing single-line strings still work
5. Test with fields that have `multiline: true` option

### DIFF
--- a/noodles-editor/src/noodles/components/field-components.tsx
+++ b/noodles-editor/src/noodles/components/field-components.tsx
@@ -24,6 +24,7 @@ import {
   type NumberField,
   Point2DField,
   Point3DField,
+  StringField,
   StringLiteralField,
   Vec2Field,
   Vec3Field,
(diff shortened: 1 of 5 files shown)


chrisgervang:
Good idea. We could just have the field component recognize the shift-enter (or presence of a new line pasted in) and switch to text area. Then all string field would support it and it's small when you don't need it.

TextLayer supports new lines, the problem is isolated to StringField's component

akre54:
I feel like there's something more fundamental going on with this and #203. Really this is about having a more Airtable-style table editor with rich text editing.

I just added #261 that pops out the editor for more advanced code. This could be similar for string fields.

But maybe the core of it is that this kind of thing is really better just handled in a CodeOp. Something like #231 could help a bunch too (see https://github.com/joby-aviation/noodles.gl/pull/203#issuecomment-3712498590)

PR #214, opened 2026-01-07
Add GraphExecutor architecture for scope-based control flow

akre54:
## Summary

- Introduces GraphExecutor with topological sorting and cycle detection for cleaner execution management
- Adds GraphScope for scope-based control flow execution (ForEach patterns)
- Adds external `compute()` function that operates on operators (not a method on Operator class)
- Adds `dirty` flag to Operator base class for change tracking
- Adds ForEachBeginOp, ForEachEndOp, ForEachMetaOp for iteration control flow (Houdini-style)
- Fields now mark their owning operator as dirty on value changes

## ForEach Operators

The new ForEach operators follow the Houdini pattern:

- **ForEachBeginOp**: Receives data array, outputs `item`, `index`, `total` - executor sets these during iteration
- **ForEachEndOp**: Collects results from each iteration into output array
- **ForEachMetaOp**: Access accumulator and iteration metadata (`initialValue`, `currentValue` → `accumulator`, `isFirst`, `isLast`)

The graph executor is responsible for managing iteration state and setting values on operators during execution. The body of the ForEach is a group node wrapping all nodes in the scope.

## Test plan

- [ ] Verify GraphExecutor creates instances correctly
- [ ] Verify topological sort produces correct order
- [ ] Verify cycle detection works
- [ ] Test ForEachBeginOp, ForEachEndOp, ForEachMetaOp creation and fields
- [ ] Verify dirty flag propagation through fields

### DIFF
--- a/noodles-editor/src/noodles/operators.ts
+++ b/noodles-editor/src/noodles/operators.ts
@@ -318,9 +318,7 @@ export abstract class Operator<OP extends IOperator> {
 
   // === Pull-based execution methods ===
 
-  /**
-   * Pull data from this operator, executing if needed (pull-based model)
-   */
+  // Pull data from this operator, executing if needed (pull-based model)
   async pull(): Promise<ExtractProps<(typeof this)['outputs']>> {
     // Return cached if clean
     if (this._pullExecutionStatus === PullExecutionStatus.CLEAN && this._cachedOutput !== null) {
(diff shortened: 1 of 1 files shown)


noodles-editor/src/noodles/graph-executor.ts
```diff
@@ -0,0 +1,938 @@
+// GraphExecutor - Execution engine for the operator graph
+// Manages operator execution with topological sorting, dirty tracking, and RAF loop
+
+import type { IOperator, Operator } from './operators'
+import { getAllOps, getOp } from './store'
+import type { OpId } from './utils/id-utils'
+
+// ForLoop operator type definitions for type checking during execution
+// These are defined here to avoid circular dependencies with operators.ts
```
chrisgervang: Type-only imports don't have runtime circular import issues since they're removed before runtime.

akre54: #227

PR #230, opened 2026-01-08
Fix for broken math pseudo operator categories

akre54:
Pseudo operators like MultiplyOp are actually just a wrapper around a MathOp with the operation type set to multiply. Previously the change in #181 broke pseudo operator categories, which made it display Unknown. We didn't have any tests so this wasn't caught, so fixing that bug now and adding some tests


akre54:
We also need to update the `headerClass` function and test it

PR #232, opened 2026-01-08
perf: Fix slow node dragging by skipping unnecessary graph transforms

akre54:
## Summary
- Node dragging was slow because `transformGraph` ran on every position change during drag (60+ times/sec)
- `transformGraph` does expensive operations (topological sorting, connection wiring) that are unnecessary for position changes
- The fix creates a stable "graph structure key" that only includes topology-relevant properties (node IDs/types, edge connections)
- The `useEffect` now depends on this key instead of `nodes`/`edges` directly, so it only runs when graph structure actually changes

## Test plan
- [ ] Start dev server and open a project with multiple nodes
- [ ] Drag nodes around - should be smooth 60fps instead of laggy
- [ ] Verify structural operations still work:
  - [ ] Add a new node (right-click → add)
  - [ ] Delete a node (select → delete)
  - [ ] Connect/disconnect edges
- [ ] Verify undo/redo still works after making changes

https://github.com/xyflow/xyflow/issues/5674#issuecomment-3854780053

### DIFF
--- a/noodles-editor/src/noodles/components/top-menu-bar.tsx
+++ b/noodles-editor/src/noodles/components/top-menu-bar.tsx
@@ -485,7 +485,6 @@ export function TopMenuBar({
                           </DropdownMenu.SubContent>
                         </DropdownMenu.Portal>
                       </DropdownMenu.Sub>
-
                     </DropdownMenu.SubContent>
                   </DropdownMenu.Portal>
                 </DropdownMenu.Sub>


noodles-editor/src/noodles/components/op-components.tsx
```diff
@@ -402,7 +389,7 @@ function OutputHandle({ id, field }: { id: string; field: Field<IField> }) {
   )
 }
 
-function NodeComponent({
+const NodeComponent = memo(function NodeComponent({
```
chrisgervang: Could we pull this standard into the style guide or a comment with a summery of the commit message explaining why this was added?

PR #240, opened 2026-01-10
Fix ForLoop to return array of all iteration results

akre54:
## Summary

- Fix ForLoopEndOp to properly collect all iteration results into an array instead of returning only the last value
- ForLoop now behaves like a map function: `[1, 2, 3]` through a ForLoop with `add 1` operation returns `[2, 3, 4]`
- Add comprehensive integration tests for ForLoop execution

## Problem

The ForLoop operators were broken after PR #214 - `ForLoopEndOp` was only returning the last value instead of collecting all iteration results. The old subscription-based `createForLoopListeners` mechanism had timing issues with async operators.

## Solution

Override `pull()` in ForLoopEndOp to:
- Iterate through the input data array from ForLoopBeginOp
- Set d/index/total on ForLoopBeginOp outputs for each iteration
- Support ForLoopMetaOp accumulator for reduce-like operations
- Mark chain operators dirty and pull each sequentially
- Collect results into an array and return them

## Test plan

- [x] All 1012 existing tests pass
- [x] Added 5 new integration tests for ForLoop execution:
  - Map-like behavior (`[1,2,3]` → `[2,3,4]`)
  - Empty array handling
  - Index/total metadata verification
  - Multiple operators in loop body
  - Object data handling
- [x] Manual UI test: Create ExpressionOp with `[1, 2, 3]` → ForLoopBeginOp → MathOp (add 1) → ForLoopEndOp, verify output is `[2, 3, 4]`

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/noodles-editor/src/noodles/operators.ts
+++ b/noodles-editor/src/noodles/operators.ts
@@ -2421,23 +2421,37 @@ export class ForLoopEndOp extends Operator<ForLoopEndOp> {
       return // No begin op found, can't set up iteration
     }
 
-    // Subscribe to the BeginOp's data input - this triggers iteration for the reactive UI path.
-    // We debounce with setTimeout(0) to allow synchronous pull() calls to complete first.
-    // This prevents race conditions between reactive execution and pull-based execution.
-    const sub = beginOp.inputs.data
-      .pipe(filter(() => !safeMode && !this.locked.value))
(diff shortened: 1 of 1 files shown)


chrisgervang:
@greptile please re-review

PR #246, opened 2026-01-12
Fix: Prevent false unsaved changes warnings on dimension updates

akre54:
#### Background

The "Leave page?" unsaved changes detection was triggering false positives when loading projects or resizing the browser window. React Flow fires 'dimensions' change events automatically whenever it measures node sizes after rendering—these were incorrectly being treated as user modifications.

#### Change List

- Filter out 'dimensions' changes alongside 'select' changes in `onNodesChange` handler
- Filter out 'dimensions' changes alongside 'select' changes in `onEdgesChange` handler
- Updated comments to clarify that we're checking for user-initiated changes only
- All 1028 existing tests pass

### DIFF
--- a/noodles-editor/src/noodles/noodles.tsx
+++ b/noodles-editor/src/noodles/noodles.tsx
@@ -222,9 +222,12 @@ export function getNoodles(): Visualization {
         analytics.track('node_selected', { count: selectedChanges.length })
       }
 
-      // Mark as unsaved if there are non-selection changes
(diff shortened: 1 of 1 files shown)


akre54:
No. But I think a user dimensions change isn't reason enough to block leaving the page

chrisgervang:
Well, it's more a question of discarding an unsaved change. I'd say a user resizing a node is a meaningful edit and therefore should put the app in an unsaved state

PR #250, opened 2026-01-12
Dim unconnectable nodes during connection drag

akre54:
#### Background
When dragging a connection from a node, it's not immediately clear which nodes can accept the connection. This makes discovering valid connection targets difficult.

#### Change List
- Add connectionDragState to UIStore to track active connection drag
- Implement onConnectStart/onConnectEnd handlers to calculate compatible nodes
- Add useNodeDimmed hook to apply dimming to incompatible nodes
- Add wrapperDimmed CSS class with opacity, grayscale, and smooth transitions
- Update all custom node components (Geocoder, Mouse, TableEditor, Viewer, Container, Time) to support dimming

https://github.com/user-attachments/assets/c52be630-47dd-4bd7-b5a7-c39569184fe5

### DIFF
--- a/noodles-editor/src/noodles/noodles.tsx
+++ b/noodles-editor/src/noodles/noodles.tsx
@@ -425,7 +425,6 @@ export function getNoodles(): Visualization {
     setConnectionDragState(null)
   }, [setConnectionDragState])
 
-
   // Hook for dropping nodes onto edges to insert them
   const { onNodeDragStop: onNodeDragStopBase } = useNodeDropOnEdge({
     getNodes: useCallback(() => nodes, [nodes]),


chrisgervang:
Nice! I think this pairs nicely with the ability to connect to anything.

What does this do when there's ambiguity in the types? Say, a data to data? I'd expect it not to be dimmed since it allows for any connection.

I do wonder if dimming is too similar to communicating the connection option "disabled" (rather than just not-necessarily-type-compatible).. would a "glow" around good connections be clearer?

akre54:
Sure that's one option! Though conceptually I'm not sure it makes a big difference. Maybe the dim could be more subtle *and* we outline connections that are good candidates

The implementation uses `canConnect` under the hood. It checks structural types - that is, a ColorField can connect to a StringField because the actual data type is a string. Same deal with DataField. Anything can connect because its underlying type is `unknown`.

In the future I'd like to add a tabular data type but for now this works

charlieforward9:
It feels like a matter of time before Noodles is used to create a bespoke animation for Noodles itself 😍💡

PR #257, opened 2026-01-12
Fix TypeScript errors and lint warnings from Zod v4 migration

akre54:
#### Background

Fixed TypeScript errors and lint warnings introduced by Zod v4 migration.

#### Change List
- Updated IOperator and IField interfaces to use any type constraints for Zod v4 compatibility
- Fixed Field subclass generic constraints (DataField, GeoJsonField, ArrayField, etc.)
- Removed unused imports and fixed unused type parameters in fields.ts and operators.ts
- Suppressed noExplicitAny lint rule for core operator/field files in biome.json
- Fixed hexToColor bug in CombineRGBAOp (removed incorrect .split() call on array)
- Fixed async execute method return types and callback parameter annotations


akre54:
I have no idea what happened here but closing

PR #258, opened 2026-01-12
fix: Silence Theatre.js cold prism warning

akre54:
#### Background

Theatre.js logs a warning 'prism.effect() does not run in cold prisms' when accessing prism values without active subscriptions. Cold prisms are those with no subscribers—effects can only run on hot prisms.

#### Change List

- Replace synchronous `val()` calls with `useVal` hook from @theatre/react
- Update `use-sheet-value.ts` to use `useVal(sheet.props)` for proper subscription
- Update `timeline-editor.tsx` to use `useVal` for sequence length
- Update `renderer.ts` to use `useVal` instead of useState + useEffect + onChange
- Reorganize `theatre-bindings.ts` to set up Theatre subscription before field subscription, caching pointer value to avoid cold prism reads
- Remove unused `val` imports from @theatre/core


akre54:
I think it didn't exist or work right or something 

Either way I think I finally understand dataverse... Now that I'm getting rid of it

PR #262, opened 2026-01-13
refactor: Use adapter pattern for Theatre field type conversions

akre54:
#### Background
Refactored Theatre.js field binding utilities to use a type adapter pattern, centralizing all Field <-> Theatre value type conversions.

#### Change List
- Created `TheatreAdapter<F, T>` interface for bidirectional field/theatre value conversion
- Implemented adapters for ColorField, DateField, Vec2/Vec3, Point2D/Point3D
- Added passthrough adapters for NumberField, BooleanField, StringField
- Refactored `fieldToTheatreProp()` and sync logic to use adapters
- Makes adding new field types easier without duplicating conversion logic


akre54:
Closed in favor of #269

PR #267, opened 2026-01-13
fix: Route-based project/example loading

akre54:
Right now both https://noodles.gl/app/projects/demo-flight and https://noodles.gl/app/examples/demo-flight point to the same project and will load the noodles file fine, but since the demo-flight is a "project" (with a file handle) and not an "example" it needs access to the /data folder. Accessing through the `/examples` route causes the FileOps in the project not to load their files from the directory. And trying to use the upload flow throws an error about it not being possible to upload to a public project.

#### Background

When accessing a user project via the `/examples` route instead of `/projects`, FileOps fail to load data files. This happens because:
1. `setActiveStorageType()` was not called after loading from user storage
2. Both routes used the same loading logic, making route semantics meaningless

#### Change List
- Fix storage type bug: add `setActiveStorageType(storageType)` after loading from user storage
- Implement route-determined source: `/examples/` only loads static bundled examples, `/projects/` only loads from user storage
- Add `ExampleNotFoundDialog` component showing navigation options when example doesn't exist
- Ensure FileOps use correct storage backend based on route

### DIFF
--- a/noodles-editor/src/noodles/noodles.tsx
+++ b/noodles-editor/src/noodles/noodles.tsx
@@ -741,7 +741,7 @@ export function getNoodles(): Visualization {
         }
       }
     })()
-  }, [projectName])
+  }, [projectName, isExamplesRoute])
 
   const displayedNodes = useMemo(() => {
     const dragHandle = `.${s.header}`


chrisgervang (commented):
Nice!

PR #269, opened 2026-01-14
Remove Theatre.js dependency

akre54:
## Summary

Comprehensive specification for replacing Theatre.js with a native timeline system in Noodles.gl.

### Documents Added

| Document | Description |
|----------|-------------|
| `README.md` | Executive summary and navigation |
| `architecture.md` | Three-tier state model, data structures, component hierarchy |
| `implementation-plan.md` | 6-phase implementation with detailed requirements |
| `ui-specification.md` | Curve editor, scrubbable inputs, timeline panel |
| `verification.md` | Side-by-side testing, interpolation accuracy |
| `research-findings.md` | Theatre.js API analysis, integration points |
| `requirements.md` | 13 requirements with acceptance criteria |

### Key Decisions

- **Option A (Full Replacement)** - Build native timeline from scratch for cleaner codebase
- **Feature flag via URL query param** - `?use_theatre=true` for gradual migration (default: native)
- **Preserve three-tier state model** - Historic (persisted), Ephemeral (session), Ahistorical (preferences)

### Notable Implementation Details

- Bezier interpolation with Newton-Raphson solver (0.0001 epsilon)
- Drag math with scrub transaction system for single-undo operations
- Type conversions: hex↔RGBA, Temporal↔epoch ms, array↔object for vectors
- Binary search for O(log n) keyframe lookup
- Zustand store with immer for immutable updates

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #272, opened 2026-01-15
Migrate render settings from Theatre.js to OutOp

akre54:
#### Background
Moves render settings management from legacy Theatre.js sheet overrides to properties of the OutOp node. This makes settings visible in the properties panel when OutOp is selected and prevents accidental keyframing.

#### Change List
- Add render settings input fields to OutOp (display, resolution, codec, bitrate, framerate, etc.)
- Create migration 012 to convert legacy settings to new format
- Add useRenderSettings() hook to read settings from OutOp
- Update TopMenuBar to select OutOp instead of opening a dialog
- Remove render settings dialog component and clean up serialization code

### DIFF
--- a/noodles-editor/src/timeline-editor.tsx
+++ b/noodles-editor/src/timeline-editor.tsx
@@ -305,7 +305,6 @@ export default function TimelineEditor() {
     })
   }, [project.address.projectId, redraw, basemapEnabled])
 
-
   // Increase the render target resolution to increase map tile detail.
   // To convert viewport bounds back to their original size, add about 1 to the zoom value.
   const lodResolution = {


noodles-editor/src/noodles/operators.ts
```diff
@@ -3609,6 +3609,18 @@ export class OutOp extends Operator<OutOp> {
   createInputs() {
     return {
       vis: new VisualizationField(),
+      // Render settings (not keyframable since OutOp is excluded from Theatre.js)
```
chrisgervang: I've wondered if keyframable should an option on the field. Defaults to true, and set to false on props like these.

I believe there are deck props that shouldn't be keyframed because they only work on initialization, but we'd have to look for them. If it only applies to this op then maybe it's not necessary as a framework feature, but it could still be nice to have.

noodles-editor/src/noodles/components/render-settings-panel.tsx
```diff
@@ -0,0 +1,300 @@
...
+            step="50"
+            onChange={e => op.inputs.captureDelay.setValue(Number(e.target.value))}
+          />
+          <span className={s.unit}>ms</span>
+        </div>
+      </div>
+
+      <button type="button" className={s.resetButton} onClick={handleResetToDefaults}>
+        Reset to Defaults
+      </button>
```
chrisgervang: Could be nice to have export photo/video buttons here too

noodles-editor/src/noodles/components/render-settings-panel.tsx
```diff
@@ -0,0 +1,300 @@
...
+function getResolutionPresetValue(width: number, height: number): string {
+  const preset = RESOLUTION_PRESETS.find(p => p.width === width && p.height === height)
+  return preset ? `${preset.width}x${preset.height}` : 'custom'
+}
+
+interface RenderSettingsPanelProps {
+  op: OutOp
+}
+
+export function RenderSettingsPanel({ op }: RenderSettingsPanelProps) {
```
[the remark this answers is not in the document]
akre54: Yeah for sure. I think it just opens us up to a little nicer / less generic property binding UI in the future too (preferably after #269).

Mostly just trying to get the presets working and easier to use than only allowing manual entry

noodles-editor/src/noodles/operators.ts
akre54: Yep! That's what #262 does. #269 could too I guess

chrisgervang: Oh that's a nice change - I was thinking of an orthogonal change of adding an option configured on the op like this:

```js
createInputs() {
  return {
    lod: new NumberField(1, { keyframable: false }) // excluded from theatre.js
  }
}
```

Does that make sense?

noodles-editor/src/noodles/store.tsx
```diff
@@ -130,12 +130,13 @@ export const useUIStore = create<UIStoreState>(set => ({
...
+  // Export functions that accept an optional OutOp ID to use for settings
+  startRender: ((outOpId?: string) => Promise<void>) | null
+  takeScreenshot: ((outOpId?: string) => Promise<void>) | null
   isRendering: boolean
   setExportActions: (actions: {
-    startRender?: () => Promise<void>
-    takeScreenshot?: () => Promise<void>
+    startRender?: (outOpId?: string) => Promise<void>
+    takeScreenshot?: (outOpId?: string) => Promise<void>
   }) => void
```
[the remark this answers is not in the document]
akre54: Actions by themselves aren't a forbidden pattern in zustand. In fact they're encouraged in some cases.

I'm open to ways of cleaning up this up further but I don't think on its own it's a code smell.

Maybe we move even more of the startRender logic to zustand so that there isn't stale closure state floating around

noodles-editor/src/noodles/operators.ts
chrisgervang: I was curious too and audited deck to find out. I'm not surprised by this, but basically all props are technically animatable. The interesting question is: which numeric props should use block keyframes rather than interpolation?

On Deck:
- _typedArrayManagerProps.poolSize - Memory pool size (integer, default: 100)
- pickingRadius - Picking tolerance in pixels (discrete pixel values)
- width, height - snap to a specific value

On DataFilterExtension:
- categorySize - Number of filter categories (int between 1-4)
- filterSize - Number of filter values (int between 1-4)

On 3D Tiles:
- pointSize - Could interpolate, but typically discrete pixel values
- loader options include a few ints

TileLayer & TerrainLayer - Tile Configuration:
 - tileSize - Pixel dimension of tiles (default: 512, usually power of 2)
 - zoomOffset - Zoom level offset (integer, default: 0)
 - maxCacheSize - Maximum number of tiles that can be cached (count)
 - maxCacheByteSize - Maximum memory in bytes for caching tiles
 - maxRequests - Maximum concurrent getTileData calls (default: 6)
 - debounceTime - Milliseconds to wait before requesting tiles (default: 0)

The list here basically consists of values I would rarely.. if ever.. keyframe, but technically could if the project called for it - the values may need special care to either be rounded, modeled as enum types, or use block keyframes. It reminds be of having "soft" vs "hard" bounds - sure, its maybe not advised but the framework doesn't need to disallow it.

So, this leads me to conclude the _only_ use case right now for `keyframable` would be on `OutOp` to prevent changing the codec during a render.

It could be nice to have an `int` field type in #269

noodles-editor/src/noodles/components/top-menu-bar.tsx
```diff
@@ -162,19 +162,60 @@ export function TopMenuBar({
...
+  const getTargetOutOpId = useCallback((): string | null => {
+    if (outOps.length === 0) {
+      return null
+    }
+    if (outOps.length === 1) {
+      // Single OutOp - use it directly
+      return outOps[0].id
+    }
+    // Multiple OutOps - check if exactly one is selected
+    const selectedNodes = reactFlow.getNodes().filter(n => n.selected)
```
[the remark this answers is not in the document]
akre54: Added active camera system. Probably more to do in the future (Blender puts this in a separate panel) but it's a good start

PR #207, opened 2026-01-05
feat: MapLibre lighting/sky operators and video rendering improvements

chrisgervang:
## Summary

This PR adds new operators for controlling MapLibre lighting and sky/atmosphere effects, plus fixes for video encoding with transparent backgrounds. 

@akre54 these should both be split out, but I'm putting this all up in case you want to make edits to the 2025 flights video which depends on these changes.

## Related PRs

This work was split out from the tools shelf feature development. The following features have been extracted into separate PRs for focused review:

### Split PRs (based off main)
- #198 - Fix Theatre.js spurious keyframes (HIGH priority)
- #199 - Add Joby color palette
- #200 - Add text rendering properties to GeoJsonLayer
- #201 - Fix GlobeView text rendering with cullMode parameter
- #202 - Add SDF font settings to TextLayer
- #203 - Add MergeOpsOp operator (depends on #198)
- #204 - Add multiline StringField with resizable StringOp
- #205 - Fix Viewer rendering for inconsistent object schemas

### Related Features
- #192 - Main tools shelf PR (Point Wizard, CSV Importer, Recipe Shelf)

### 1. MapLibre Light and Sky Operators

**MaplibreLightOp** - Control directional lighting without reloading the map style:
- `anchor`: 'map' or 'viewport' positioning
- `azimuthal`: Azimuthal angle (animatable, no constraints)
- `polar`: Polar angle (animatable, no constraints)
- `radial`: Radial distance (animatable, no constraints)
- `color`: Light color
- `intensity`: Light intensity (0-1)

**MaplibreSkyOp** - Control sky and atmosphere rendering:
- `skyColor`: Sky color (default '#80a5ff')
- `horizonColor`: Horizon color (default '#ffffff')
- `fogColor`: Fog color (default '#ffffff')
- `fogGroundBlend`: Fog ground blend (0-1)
- `atmosphereBlend`: Atmosphere blend (0-1)

**MaplibreBasemapOp** - Updated to accept optional `light` and `sky` inputs

**Implementation Details:**
- Uses imperative `map.setLight()` and `map.setSky()` APIs to avoid style reloading
- Prevents flashing during animation when lighting/sky parameters change
- Added to 'view' category in operator menu

### 2. Video Rendering Fixes

**Alpha Channel Compositing:**
- Creates intermediate canvas to composite transparent frames onto black background
- Solves corruption issues with codecs that don't support alpha (H.264, H.265, VP9, AV1)
- Uses `alpha: false` context option for optimal performance
[...]

PR #281, opened 2026-01-22
feat: Add sky and lighting settings to MaplibreBasemapOp

chrisgervang:
## Summary
- Add sky and light settings directly to MaplibreBasemapOp instead of separate operators (replaces #207)
- Sky settings (optional): skyColor, horizonColor, skyHorizonBlend, atmosphereBlend
- Light settings (applied to globe when using sky): anchor, azimuthal, polar
- Increase pitch max from 60° to 85°

Screenshot 2026-01-22 at 11 49 58 AM

### Globe Atmosphere
nyc-taxis-globe-light

nyc-taxis-atmosphere

### Sky

nyc-taxis-sky

## Implementation Details
- Use imperative `map.setLight()` and `map.setSky()` to avoid style reloading/flashing during animation
- Sky colors apply to all projections at certain zoom/pitches; light and atmosphere apply to globe projection
- Disabling sky uses `setSky(undefined)` (requires MapLibre GL JS 4.6.0+)

## Test plan
- [x] Open a project with MaplibreBasemapOp
- [x] Enable sky, adjust colors - verify no map flashing
- [x] Switch to globe projection, adjust light azimuthal/polar - verify lighting changes
- [x] Disable sky - verify it resets
- [x] Test pitch up to 85°

### DIFF
--- a/noodles-editor/src/noodles/operators.ts
+++ b/noodles-editor/src/noodles/operators.ts
@@ -3278,18 +3278,24 @@ export class MaplibreBasemapOp extends Operator<MaplibreBasemapOp> {
         pitch: new NumberField(0, { min: 0, max: 85, optional: true }),
         bearing: new NumberField(0, { optional: true }),
       }),
-      sky: new CompoundPropsField({
-        enabled: new BooleanField(false),
-        skyColor: new ColorField('#88C6FC'),
-        horizonColor: new ColorField('#ffffff'),
-        skyHorizonBlend: new NumberField(0.8, { min: 0, max: 1, step: 0.01 }),
-        atmosphereBlend: new NumberField(0.5, { min: 0, max: 1, step: 0.01 }),
-      }),
-      light: new CompoundPropsField({
-        anchor: new StringLiteralField('viewport', ['map', 'viewport']),
-        azimuthal: new NumberField(210, { step: 1 }),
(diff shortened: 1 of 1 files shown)


chrisgervang:
https://github.com/user-attachments/assets/eb6982cd-2f84-45d3-8658-2db080c96859

We could make the fields list editable. Every op starts with a small set of defaults and the user can just add what they need. This was a quick prototype, but imagine the fields on the node dynamically showing/hiding as the user edits this.

chrisgervang:
Yay! Yes, that came to mind - I bet they'd share logic around setting up and tearing down subscriptions. I think they're different use cases too.. a parameter editor is soft-coded to whatever field you want, and these fields are hard-coded and curated by Noodles for the op. Both good to have

I think its a worthy tradeoff, and we could also put guards in place - an "are you sure?" dialog or disable removing a field if the field has a connection. And some inputs shouldn't be removable too.. `index` on SwitchOp comes to mind, viewState on `maplibre` too. Libraries tend to document their minimally required props

`extras` or adding additional ops are less discoverable (user needs to learn "related" ops) and could still come with fields the user doesn't want to see. MergeOp-like solutions also ballon the number of ops a user needs (what could be done with 1 now requires 4 op - basemap, sun, sky, merge). I think this idea addresses all of these well

PR #286, opened 2026-01-24
feat(operators): add controller configuration to DeckRendererOp and Views

charlieforward9:
## Summary
- Add configurable controller options for deck.gl interactivity to DeckRendererOp
- Add controller BooleanField to View operators (MapViewOp, GlobeViewOp, etc.)

### DeckRendererOp controller options:
- scrollZoom, dragPan, dragRotate, doubleClickZoom
- touchZoom, touchRotate, keyboard, inertia
- Defaults to `true` (all interactions enabled) when not configured

### View operators:
- Add `controller` BooleanField to enable/disable user interaction per view

## Test plan
- [ ] Verify controller options appear in DeckRendererOp node
- [ ] Test disabling individual interactions (e.g., scrollZoom: false)
- [ ] Test controller toggle on View operators

### DIFF
--- a/noodles-editor/src/noodles/operators.test.ts
+++ b/noodles-editor/src/noodles/operators.test.ts
@@ -807,9 +807,20 @@ describe('DeckRendererOp', () => {
     expect(mapProps?.mapStyle).toBe('')
   })
 
+  it('omits controller from deckProps by default', () => {
+    const operator = new DeckRendererOp('/deck-0')
+    const {
+      vis: { deckProps },
+    } = operator.execute({
+      layers: [],
+      effects: [],
+      views: [],
+      layerFilter: () => true,
+    })
+    expect(deckProps).not.toHaveProperty('controller')
+  })
+
   it('includes basemap viewState in deckProps when mapStyle is empty (transparent mode)', () => {
-    // When mapStyle is empty, timeline-editor switches to standalone DeckGL (basemapEnabled=false).
-    // deckProps.viewState must carry the geo position so layers render at the correct location.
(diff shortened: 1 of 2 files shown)


akre54 (commented):
Thanks @charlieforward9 !

Can you please think through some example cases and create a test for the functionality you want to support with this? We have the `showOverlay` editor setting that may help (see also #8)

chrisgervang (commented):
Let's dig into the use case a bit more to be confident it meets your need while also being disabled for projects that don't need interactively

Also something to keep in mind is these props are always disabled when using a base map (See https://deck.gl/docs/api-reference/mapbox/mapbox-overlay#constructor) since interactivity is delegated to the map library, so if your plan was to use maplibre we should look at `MaplibreBasemapOp` for potentially exposing what your use case calls for

charlieforward9:
Addressed all review feedback:

- **Simplified to BooleanField** — removed CompoundPropsField, just a single boolean toggle for controller
- **Defaults to disabled** (false) to match deck.gl defaults and avoid overhead
- **Omits controller key** from deck props unless explicitly enabled (uses spread: `...(controller ? { controller } : {})`)
- **View controller** also defaults to false with `showByDefault: false`, matching the DeckRendererOp API
- **Added tests** for DeckRendererOp controller (omit by default, include when enabled, omit when false) and MapViewOp controller behavior

charlieforward9 (commented):
This should have no negative backwards compatibility issues to merge , correct?

charlieforward9:
I can appreciate that. 

Ill close this in its favor.

PR #287, opened 2026-01-24
feat(TextLayerOp): add depthTest and depthWriteEnabled parameters

charlieforward9:
## Summary
- Add WebGL depth buffer parameters to TextLayerOp for better control over text rendering in 3D scenes

### New parameters:
- `depthTest`: Enable/disable depth testing (default: true)
- `depthWriteEnabled`: Enable/disable depth buffer writes (default: true)

## Motivation
These parameters are useful when rendering text on GlobeView or in scenes with overlapping 3D elements where depth ordering matters.

## Test plan
- [ ] Verify parameters appear in TextLayerOp node
- [ ] Test with GlobeView to confirm proper text rendering
- [ ] Test toggling depthTest off for overlay text effects


akre54:
Yes exactly. #35 needs some thinking through but that's probably the direction we'll go

PR #295, opened 2026-01-28
fix: Find shallowest noodles.json when importing ZIP files

### LINKED ISSUE #293: Import flow creates new project template at root, embeds imported project in subfolder
## Description

When importing a project ZIP file, the import flow incorrectly writes a new project template `noodles.json` at the root of the destination folder, and the actual imported project files end up in a subfolder with the original project name.

## Steps to Reproduce

1. Fresh page load
2. Navigate to an example (e.g., `/examples/nyc-taxis`)
3. File > Import
4. Select a project ZIP file
5. In the folder picker, click "New Folder", name it (e.g., "my-import"), and select it

## Expected Behavior

- Destination folder contains only the imported project files
- `noodles.json` at root should be the imported project data
- Data files should be at the expected paths (e.g., `data/file.csv`)

## Actual Behavior

- Root `noodles.json` contains the **new project template** (DeckRendererOp, OutOp, MaplibreBasemapOp)
- Imported project files are nested in a subfolder with the original project name

## Environment

- Occurs in both development and production
- Browser: Chrome (File System Access API)
- OS: macOS

## Root Cause Hypothesis

Something in the import flow is triggering `onNewProject` in addition to (or instead of) the import file writing logic. The new project template is only written by `onNewProject` in `noodles.tsx`.

## Related

This was discovered while testing PR #275 (fix: Navigate to /projects route when opening user projects).

chrisgervang:
Closes #293

#### Background
When importing a ZIP file that contains multiple `noodles.json` files (e.g., a project's main `noodles.json` and another used as dummy data in a subfolder), the import would use the wrong `noodles.json` as the project root, causing imported files to be nested incorrectly.

#### Change List
- Changed ZIP extraction to find the shallowest `noodles.json` (fewest path segments) instead of the last one found
- This ensures the project root is correctly identified even when nested `noodles.json` files exist
- All existing tests pass

### DIFF
--- a/noodles-editor/src/noodles/noodles.tsx
+++ b/noodles-editor/src/noodles/noodles.tsx
@@ -1095,16 +1095,25 @@ export function getNoodles(): Visualization {
         // Handle ZIP import
         const zip = await JSZip.loadAsync(await file.arrayBuffer())
 
-        // Find noodles.json in the ZIP (could be at root or in a subfolder)
+        // Find the shallowest noodles.json in the ZIP (could be at root or in a subfolder)
+        // We want the project root, not a noodles.json that might be in a data subfolder
         let noodlesJsonPath: string | null = null
         let projectFolder = ''
 
-        // Iterate over files in the ZIP to find noodles.json
+        // Iterate over files in the ZIP to find the shallowest noodles.json
         zip.forEach((relativePath, zipEntry) => {
           if (relativePath.endsWith('noodles.json') && !zipEntry.dir) {
-            noodlesJsonPath = relativePath
             // Extract the folder path (everything before noodles.json)
-            projectFolder = relativePath.substring(0, relativePath.lastIndexOf('noodles.json'))
+            const folder = relativePath.substring(0, relativePath.lastIndexOf('noodles.json'))
+            // Keep this one if it's the first found, or if it's shallower (fewer path segments)
+            const currentDepth = noodlesJsonPath
+              ? projectFolder.split('/').filter(Boolean).length
+              : Infinity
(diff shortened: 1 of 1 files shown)


akre54:
Oh - seems like a corner case but I guess good to add. Worth adding a test?

chrisgervang:
It's on my todos to split this stuff out of noodles.tsx.. that file isn't testable

akre54:
It's tough because it's really the integration point of so many different parts. I'm not sure just splitting it apart fixes the integration pain point, even if all together it's got a lot going on

chrisgervang:
No, totally. I was referring to project management - which would still integrate in this file, but seems like it can be its own module/store.. tbd

akre54:
> but seems like it can be its own module/store.. tbd

Ok curious your thoughts when you have 'em. In my eyes it's not a huge deal whether it lives in ReactFlow's store or our zustand store but maybe I'm missing something

charlieforward9:
Powerful deep dive. 

Im able to utilize the pieces of the system that I need easily at this point - so I am indifferent to standardizing things more than they already are. Just adding to the conversation with a lil POV.

However, for the use case I am designing for - I use all of noodles headlessly, so any further integration of UI and data model would negatively impact my implementation. Noodles manages my state, but the user doesn't know its there - kinda like zustand or redux would work.

akre54:
Totally understand and thanks for the perspective. And if it's working for you that's great. It's definitely possible we carve off a few of the more free-floating parts (like a timeline component, say) into a separate package or repository if other use cases call for it.

PR #302, opened 2026-02-10
feat: Add createFromSQL query parameter for projects

akre54:
Add a new `createFromSQL` query parameter that creates a new project with a pre-configured DuckDBOp containing the SQL query from the parameter. The generated graph includes ScatterplotLayerOp, AccessorOps for position and color, and all standard rendering nodes.

Changes:
- Extract SQL validation and graph generation into pure function (create-from-sql.ts)
- Intercept at onNewProject() to check for and process createFromSQL parameter
- Remove query param from URL after project creation
- Add 16 unit tests for graph generation and SQL validation
- Add 9 integration tests verifying operator creation and connections
- Track analytics with method: 'createFromSQL'

All tests passing (25/25 ✓) and lint clean.


charlieforward9 (commented):
Would be cool to customize the example data onboarding modal to work like a 2 or 3 tiered select menu so it adjusted: 

SELECT (output opts) FROM (source opts) WHERE (filter opts)

PR #305, opened 2026-02-26
fix: Skip type checking on ReferenceEdges

akre54:
## Background

ReferenceEdges represent reactive operator references (e.g., in code expressions via `{{/operator.out.value}}`), not data flow connections. Type validation was incorrectly applied to them, causing false "Type mismatch" errors when connecting operators with incompatible field types via reference.

For example, connecting a NumberOp to a CodeOp's code parameter would produce: "Type mismatch: number cannot connect to code."

## Change List

- Skip type validation for ReferenceEdges in `transform-graph.ts` (only validate value edges)
- Added test: ReferenceEdges do not produce type mismatch errors
- Added test: value edges with type mismatches still produce errors (regression guard)

Fixes type mismatch false positives while maintaining validation for actual data flow connections.

### DIFF
--- a/noodles-editor/src/noodles/transform-graph.test.ts
+++ b/noodles-editor/src/noodles/transform-graph.test.ts
@@ -1,7 +1,14 @@
 import type { Node as ReactFlowNode } from '@xyflow/react'
 import { afterEach, describe, expect, it } from 'vitest'
 import type { Edge } from './noodles'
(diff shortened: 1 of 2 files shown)


chrisgervang (commented):
I look forward to using the less noisy logs!

akre54:
It's less about the logs - unfortunately we weren't logging anything previously

ReferenceEdges were just broken, and we didn't have a way of spotting that

PR #313, opened 2026-02-27
feat: add /projects page with modal routes and File menu links

akre54:
Add a new /projects route that displays user-saved projects from FileSystemAccess API handles (cached in IndexedDB) and OPFS directories. Both /projects and /examples routes now render as full-screen modal overlays over the app with a close button and navigation strip, allowing users to easily dismiss them via ESC key, back button, or the X icon. Add "View Recent..." menu item to File > Open Recent that navigates to /projects, and add "Examples" link to File menu. In dev mode, the app defaults to /projects; in prod mode, it defaults to /examples.


akre54:
It's ugly but it's a start. Merging

PR #314, opened 2026-02-27
fix: prevent stale frame from being encoded at render start

akre54:
#### Background

When starting a video render with the timeline playhead at a non-zero position, the first captured frame could contain stale visual content from the previous playhead position. This happened because the capture loop set the playhead position but the canvas buffer still contained content from before the render completed.

#### Change List

- Add warmup render cycle before capture loop to ensure canvas contains correct frame content
- Seek to start frame, tick Theatre.js, and wait for render completion before entering capture loop
- Remove outdated TODO comment about first frame capture


chrisgervang (approved):
🙏

PR #316, opened 2026-02-28
fix: prevent memory leak during video export

akre54:
## Summary

Fixes video export crashes when timeline exceeds ~3.5s. The issue was a debug line passing large WebGL objects to console.warn(), which React DevTools retained as 52.9MB strings in its console history — 68 frames × 52.9MB ≈ 3.6GB.

## Changes

- Removed console.warn('redraw', mapRef.current, deckRef.current) — was serializing WebGL device state to console
- Gated setRand(Math.random()) behind !isRendering to prevent React re-renders during export (was growing CSSStyleRule/DOM)
- Removed per-frame console.log calls in renderer loop

🤖 Generated with Claude Code


akre54:
Wow this might be the dumbest reason for a memory leak / performance regression.

React DevTools holds onto console logs for some reason. So that passing the Deck ref to the `console.warn` caused an explosion of `JSON.stringify` memory leaks. Yuck

PR #318, opened 2026-03-01
feat: Enable sourcemaps with R2 storage and automatic cleanup

akre54:
## Summary

Enable sourcemap generation in production builds, upload them to Cloudflare R2 keyed by git SHA, and automatically clean up old maps after 60 days. Sourcemaps are removed from the GitHub Pages artifact to prevent deploy failures.

## Change List

- Enable sourcemap generation in Vite build
- Add upload-sourcemaps.ts script to rewrite sourceMappingURL comments with R2 URLs (keyed by SHA), upload .map files to R2, and strip .map files from dist/ before GitHub Pages upload
- Add cleanup-sourcemaps.ts script to delete maps older than 60 days that are not referenced by active staging environments
- Update deploy-docs.yml to split build into three steps: build app → upload maps to R2 → build website; ensures .map files never reach the Pages artifact
- Add cleanup-sourcemaps.yml workflow for weekly automatic cleanup (Sundays 03:00 UTC) plus manual dispatch
- Add upload-sourcemaps and cleanup-sourcemaps npm scripts


akre54:
Thanks for the review! Both issues have been fixed in the follow-up commit (b2bfcca):

- **Pagination bug**: Changed `--starting-token` to `--continuation-token` in `cleanup-sourcemaps.ts`. `--starting-token` is a CLI-level flag for `--max-items` flow and doesn't map to the S3 `ContinuationToken` API parameter — the correct flag for passing `NextContinuationToken` back to `list-objects-v2` is `--continuation-token`.

- **CF Pages build failure**: Added `sourcemap: !env.CF_PAGES` in `vite.config.js`. CF Pages runs its own build without our R2 upload step, so maps would bloat/break that artifact. CF Pages automatically sets `CF_PAGES=1`, so sourcemaps are only generated in the GitHub Actions deploy (where they get uploaded to R2 and stripped from `dist/` before the Pages artifact is assembled).

PR #378, opened 2026-03-24
feat: Add MapStyleConfigurator operator for visual map style editing

akre54:
#### Background

Adds a new `MapStyleConfigurator` operator for visually editing Maplibre/Mapbox GL style properties without manually editing JSON or style URLs. Users can adjust layer colors, toggle visibility, scale label sizes, and override the glyphs URL through a dedicated dialog UI. Overrides are stored as sparse JSON and merged into the base style at execution time, with live preview as changes are made.

#### Change List
- Add `MapStyleConfiguratorOp` operator with `baseStyle` (URL or inline JSON) and `overrides` inputs; outputs a merged style JSON connectable to `MaplibreBasemapOp`
- Add `map-style-configurator-op.tsx` with `MapStyleConfiguratorDialog` — a Radix UI dialog grouping layers into semantic categories (Background, Labels, Water, Land, Buildings, Roads, Other) with per-layer color swatches, visibility toggles, and global label size scale + glyphs URL overrides
- Add `src/utils/map-style-utils.ts` with pure utility functions (`categorizeLayer`, `applyStyleOverrides`, `groupLayersByCategory`) and 20 unit tests
- Export shared node helpers (`useLocked`, `useNodeDimmed`, `useFieldVisibility`, `NodeHeader`, `OutputHandle`) from `op-components.tsx`; register operator in the `view` category


akre54:
> maybe add a map preview

Definitely. That was an early thought but I think you're right it makes sense to include. Will add

> I'm imagining we can add toolbar tools to help a user "build" a custom op out of known formats like this

Yes!

Eventually lots of recipes, or the equivalent of [TouchDesigner's Palette](https://derivative.ca/UserGuide/Palette)

> We'd need to build our own JSON-delta-to-imperative-API engine to do this downstream of the graph.

Yeah https://acko.net/blog/reconcile-all-the-things/ has been a big inspiration for a lot of how I think about Noodles - basically all output is a descriptor and is lazily evaluated *except for* the intermediate code transforms, which are comparatively quick. Eventually I want it all to sorta work like this

PR #393, opened 2026-03-27
refactor: replace MapStyleOp with typeahead on MaplibreBasemapOp

akre54:
#### Background

`MapStyleOp` existed solely as an intermediary to expose preset CARTO map styles as a dropdown, requiring users to wire it into `MaplibreBasemapOp`. This simplifies the pattern by removing that node and surfacing the presets directly on the basemap node itself.

#### Change List
- Add `suggestions` option to `FileUrlField` (and store on the instance) so any file-url input can show typeahead hints via an HTML5 `<datalist>`
- Update `FileUrlFieldComponent` to render the datalist when suggestions are present
- Update `MaplibreBasemapOp.inputs.mapStyle` to include the `MAP_STYLES` presets as suggestions
- Remove `MapStyleOp` class, its `opTypes` registration, and its category entry


chrisgervang (approved):
Nice UX, agreed this didn't need to be an op on its own

PR #396, opened 2026-03-30
feat: add free API fallbacks and key nudges

akre54:
#### Background

Several operators hard-failed when API keys weren't configured. This adds free fallbacks and surfaces nudges so users understand what keys unlock.

#### Change List
- **DirectionsOp**: driving mode now falls back to OSRM (free, OpenStreetMap-based routing) when no Mapbox token is set; fails with a helpful error only if OSRM also fails
- **Geocoding dialog**: shows a provider badge after each place search indicating Photon/Mapbox/Google Places, with a Settings link nudge when using the free Photon fallback
- **Settings dialog**: updated all four API key descriptions to explain what each key enables and what free alternatives exist without it


charlieforward9:
https://noodles.gl/app/examples/cesium-hubble throws `[Noodles] Operator execution error: /renderer Error: Tile3DLayer: Cesium Ion access token is not set (add it in Settings > API Keys)` in console but has no UI visibility - is this included in here?

akre54:
I don't think we have a good solution to Cesium Ion. What do you propose

charlieforward9:
@akre54 been trying to see this demo for weeks haha

Is there no way for you to create an api key scoped exclusively to noodles.gl? I have never created a Cesium account 

this demo was going to be my motivation

PR #414, opened 2026-04-11
fix(operators): preserve extension props in parseLayerProps output

charlieforward9:
## Summary
- `parseLayerProps` was stripping extension constructor args by mapping `validExtensions.map(e => e.extension)`, discarding the `props` object (e.g. `{ terrainDrawMode: 'offset' }`)
- Downstream instantiation received `{ type: 'TerrainExtension' }` with no constructor args, causing extensions like `_TerrainExtension` to fall back to defaults instead of using the nib-declared mode
- Keep the full `{ extension, props }` structure so consumers can pass both the extension type and its constructor args

## Test plan
- [x] Verify layers with `TerrainExtension` declared in nib `data.inputs.extensions` correctly drape on terrain
- [ ] Verify extension props (e.g. `terrainDrawMode: 'offset'`) are passed to extension constructors
- [ ] Verify backward compatibility with existing projects using extensions

### DIFF
--- a/noodles-editor/src/noodles/operators.ts
+++ b/noodles-editor/src/noodles/operators.ts
@@ -4193,10 +4193,12 @@ function parseLayerProps<P extends LayerProps>({
   const result = { ...props }
 
   if (validExtensions.length > 0) {
-    // Keep extensions as POJOs for now - they'll be instantiated later in noodles.tsx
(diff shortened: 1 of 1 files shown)


akre54:
Add tests that fail before this change but pass after?

PR #447, opened 2026-05-04
fix: reduce maplibre flickering with deep equality checks and icon caching

akre54:
#### Background

MapLibre basemap was flickering during timeline playback because operators were creating new object references on every execution, triggering unnecessary field updates and map style reloads even when content was identical.

#### Change List
- Add `deepEqual` utility in standalone file for deep object comparison
- Implement deep equality checking in `MapStyleField.setValue()` to skip updates when object content is identical
- Extend operator output comparison to use deep equality for `CompoundPropsField` and object-valued `MapStyleField` outputs
- Add icon caching to `IconLayerOp` with LRU eviction to prevent texture re-uploads and flicker
- Add comprehensive tests: 196 tests for `deepEqual`, 87 tests for `MapStyleField` deep equality, 250 tests for operator output comparison
- Remove unused `memoize.ts` and `memoize.test.ts` files


noodles-editor/src/noodles/noodles.tsx
```diff
@@ -1596,7 +1596,19 @@ export function getNoodles(): Visualization {
   const OUT_OP_ID = '/out'
   const outOp = operators.find(n => n.id === OUT_OP_ID)! as unknown as OutOp
 
-  const [visProps, setVisProps] = useState(outOp?.inputs.vis.value || {})
+  // Find the DeckRendererOp that connects to OutOp
+  // Subscribe to its output instead of OutOp's input to create active subscription chain
+  const deckRendererOp = useMemo(() => {
```
chrisgervang, line 1655: What is the benefit of subscribing to teh output instead of the input? Are they not the same, or is deep equal only applying downstream?

noodles-editor/src/noodles/operators.ts
```diff
@@ -6942,7 +6958,7 @@ export class SimplifyOp extends Operator<SimplifyOp> {
   createInputs() {
     return {
       feature: new GeoJsonField(),
-      tolerance: new NumberField(0.01, { min: 0.0001, softMax: 1, step: 0.01 }),
```
chrisgervang, line 6985: Simplify tolerance can go negative?

chrisgervang (approved):

noodles-editor/src/noodles/operators.ts
akre54: I guess it should be 0

PR #452, opened 2026-05-05
perf: use MapLibre render event for 8.6x faster video export

akre54:
## Summary

- Replace MapLibre's `onIdle` event (300ms internal debounce) with the `render` event during export, achieving near-realtime frame capture speed
- Implement "skip first render, capture on second" strategy to avoid stale/intermediate frames
- Reduce pure-deck export delay from configurable `captureDelay` to minimal 16ms safety margin
- Add performance regression tests to guard against future export speed degradation

## Performance Results

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| Frame time | 308ms | 36ms | **8.6x faster** |
| Export FPS | 3.2 | 27.8 | 8.7x |
| Realtime factor | 0.11x | 0.93x | Near-realtime |
| Wait phase (% of frame) | 98.7% (304ms) | ~44% (16ms) | Eliminated debounce |
| Capture + Encode | <1ms | <1ms | Unchanged |

Test scene: geospatial trace visualization with MapLibre basemap, cached tiles, timeline-driven animation.

## How It Works

MapLibre's `onIdle` fires after an internal debounce (~300ms) designed for interactive use — coalescing multiple map events. During sequential frame export this is pure waste since we control exactly when each frame is requested.

The new approach:
1. `redraw()` resets a frame counter and `frameCaptured` guard
2. MapLibre's `render` event fires immediately after each render pass (~13-23ms)
3. We skip the first render (may show stale operator data from previous timeline position)
4. On the second render, operators have settled — capture the frame with a 16ms safety margin for GPU buffer swap
5. `onIdle` remains as a fallback (in case `render` misses due to edge cases) and as the primary handler during interactive preview

## Alternatives Considered

| Approach | Result | Why rejected |
|----------|--------|--------------|
| Immediate capture on first `render` | Stale/flickering frames | First render after `redraw()` may show previous frame's operator data |
| `setTimeout(0)` after render | Still captures first render | Race condition with operator execution |
| `requestAnimationFrame` after render | Tab-throttled to 1fps by Chrome | Broken when tab is backgrounded |
| `onRender` as React prop on `<ReactMapGL>` | Continuous repaints in interactive mode | Can't scope to export-only without `useEffect` |
| `map.once('idle')` per frame | Same 300ms debounce | Fundamental limitation of idle detection |

## Potential Pitfalls

1. **First-frame stale data**: If the "skip first" heuristic is wrong for some scene types, frames could be captured too early. The `onIdle` fallback handles this case (fires after everything truly settles).
2. **Scenes with heavy async data loading**: The `waitForData` + `deck.props.layers.every(l => l.isLoaded)` check ensures we wait for async data, but very slow data sources could still cause the `render` event to fire many times before data arrives.
3. **`useEffect` dependency on `isRendering`**: The render event listener mounts/unmounts based on React state — there's a brief window during state transition where neither handler is active. The warmup frame in `renderer.ts` covers this gap.
4. **16ms safety margin**: Chosen as one RAF tick for GPU buffer swap. If a GPU is extremely slow to present, this could theoretically be insufficient, but in practice the compositor read (`captureStream(0)`) waits for the presented buffer.

## Test Plan

- [x] All 10 existing export performance tests pass
- [x] New regression tests validate:
  - Render-event approach achieves >5x speedup over onIdle
  - Cached-tile scenes stay above 0.5x realtime
  - Skip-first-render strategy correctly guards against stale frames
- [ ] Manual test: export 100 frames from `nyc-april-2026-traces` project, visually verify no blank/flickering frames
- [ ] Manual test: interactive preview still works normally (no flickering on basemap)
- [ ] Manual test: export with tab backgrounded still completes (worker timer drives render pump)


chrisgervang (changes requested):
This hasn't worked in the past, and subtly breaks renders.

The core problem is that the render event doesn't guarantee to be the "settled" state of a frame - there are many intermediate draw calls so it can move on to the next frame before the current frame is actually finished.

Unless something has fundamentally changed in how maplibre or the renderer works, I don't recommend moving forward with this. It's been tried multiple times and was unsuccessful

chrisgervang:
It certainly would be nice to have these render gains. I'd just like to be very cautious and detailed in how we validate this new approach actually works.

I think it needs to be tried on multiple projects and golden image render tested so we can know for certain the videos render consistently before/after

akre54:
Thanks for the review @chrisgervang. Yeah this is a work-in-progress and it depends on #447 but I saw some promise when they were in the same branch so maybe there's an issue with how we were doing operator memoization. The speedups alone make this worth exploring, and we can add a checkbox if the render errors

Do you remember the specifics of why we needed to wait for the idle callback? That should be more about loading data than it is about drawing the framebuffer right?

Our renderer pipeline is super complicated and has a bunch of critical signals. Worst of all, it's really under-tested. So I'm hoping that little by little we can chip away at this and add automated testing so this stuff doesn't break

chrisgervang:
> Do you remember the specifics of why we needed to wait for the idle callback? That should be more about loading data than it is about drawing the framebuffer right?

My memory on this is vague, but I remember writing up PR descriptions in Hubble and a noodles repo about specific issues the idle event solved over the years.

I don't think it's directly related to data loading, though that plays a role. I think it's more about determining "when has the frame settled?" to signal when it's time to capture.

A symptom of capturing too soon is repeated frames (a frame signaled it was ready too early, but visually, the frame still looked the same).

The "render" draw fires early and multiple times during a frame's preparation - it'll fire when a style is updated, when I tile is loaded, again when another tile is loaded, when a hover interaction occurs, etc.. 

Eventually, a render event fires when it's an appropriate time to capture the frame - but the problem is knowing which render event will be the _last_ one?

Idle, on the other hand, is designed to fire after a period of inactivity.. I'll send a summary of how it's implemented. But basically it's singling "the render loop has nothing to do" which is a lot closer to signaling the _last_ draw in a frame's preparation.

chrisgervang:
From CC:

Here's how it works in MapLibre GL JS. The whole thing is ~10 lines at the bottom of `_render()` in [src/ui/map.ts](https://github.com/maplibre/maplibre-gl-js/blob/main/src/ui/map.ts).

## The render loop drives it

Everything flows through `_render(paintStartTimeStamp)`. It's invoked from a `requestAnimationFrame` callback scheduled by `triggerRepaint()`:

```ts
// map.ts:~3819
triggerRepaint(): void {
    if (this.style && !this._frameRequest) {
        this._frameRequest = new AbortController();
        browser.frame(this._frameRequest, (paintStartTimeStamp) => {
            this._frameRequest = null;
            try { this._render(paintStartTimeStamp); }
            ...
```

The `if (!this._frameRequest)` guard collapses many `triggerRepaint()` calls into a single frame — that's why callers can spam it.

## The idle decision

At the very end of `_render()` (map.ts:~3725):

```ts
const somethingDirty = this._sourcesDirty || this._styleDirty || this._placementDirty;
if (somethingDirty || this._repaint) {
    this.triggerRepaint();                 // schedule another frame
} else if (!this.isMoving() && this.loaded()) {
    this.fire(new Event('idle'));          // we're done
}
```

So `idle` is **the absence of work**. It fires when, after a paint:
1. No dirty flags are set (style, sources, placement),
2. The camera isn't animating (`!isMoving()`),
3. `loaded()` returns true.

If any of those fail, the map either schedules another frame or simply stops rendering (and `idle` won't fire until the next render that lands in the clean branch).

## What `loaded()` means

```ts
loaded(): boolean {
    return !this._styleDirty && !this._sourcesDirty
        && !!this.style && this.style.loaded();
}
```

`style.loaded()` walks every source's `TileManager` and checks `areTilesLoaded()` plus pending sprite/glyph requests. So `idle` transitively requires every visible tile, every glyph atlas, and every image to have arrived — not just "no animation in progress."

## How the dirty flags get set

This is the other half of the picture — what *prevents* `idle`:

- `_update(updateStyle?)` sets `_sourcesDirty = true` (and optionally `_styleDirty`) and calls `triggerRepaint()`. Called from `setPaintProperty`, source data events, etc.
- `_placementDirty` is the return value of `style._updatePlacement(...)` — symbol fade/collision work in progress.
- `style.hasTransitions() || crossFading` re-arms `_styleDirty` mid-render so a transition keeps frames coming.
- Camera animations call `_update()` each frame via the `_renderTaskQueue`.

So the loop self-perpetuates as long as anything wants more frames; once nothing does, the next render hits the `else` branch and emits `idle`.

## The `_idleTriggered` flag

In the constructor (map.ts:810):

```ts
this.once('idle', () => this._idleTriggered = true);
```

It's used at the top of `_render()`:

```ts
const fadeDuration = this._idleTriggered ? this._fadeDuration : 0;
```

This suppresses symbol cross-fading until the *first* idle. Initial load shouldn't fade — only later updates should. Unrelated to firing `idle` itself, but it's what that flag is for.

## Mental model

`idle` is not a one-shot "load complete" signal — it fires every time the map settles. Pan the map: `idle` → camera moves → many renders → settles → `idle` again. It's effectively "the render loop has nothing to do."

Two related events for contrast:
- `load` — fires *once*, also from inside `_render`, the first time `loaded()` flips true (map.ts:3707).
- `render` — fires *every* `_render()` call, before the idle check.

akre54:
Totally understand, and thanks for the deep dive. I just feel like waiting on every frame on the off chance that data has changed is *really* wasteful / slow. It makes every render take an eternity, which is especially tough with some of these longer-form videos when we're making minor tweaks (and there isn't a ton of complexity on the map area).

I wonder if we can hook more deeply into the maplibre dirty checking, or if we should do our own tracking of inputs that go into maplibre (i.e. if we update the mapStyle). Or if we should insert an idle check every x frames. I just really want exports to be quick, and right now we're getting about 0.1 fps in renders

chrisgervang:
Where does it say it's waiting on the off-chance something isn't loaded? There isn't a timer or sleep in maplibre's idle event

It sounds like you've got a need for fast renders during high iteration period where you'd care less about rendering correctly and more about rendering speed.. that's valid. 

I think that's a different mode from rendering a final where the output needs to be perfect. We've optimized for the latter

Many tools come with a scratch renderer.. maybe we need one too

akre54:
> It sounds like you've got a need for fast renders during high iteration period where you'd care less about rendering correctly and more about rendering speed.. that's valid.

Well kinda. It's both. I don't think we really have to sacrifice quality. It plays back just fine in realtime when I hit the spacebar. And I don't want to do a bunch of iteration just to find out that the output doesn't match. We don't need a separate renderer for that

The issue is that I can't just send over scratch video. The times when I have the most notes are later on in the process, and that's when deadlines are tightest so I need the fastest iterations. 

> I don't think it's directly related to data loading, though that plays a role. I think it's more about determining "when has the frame settled?" to signal when it's time to capture.

I guess I'm saying it should theoretically be more obvious / earlier than the idle event. Again, something to explore. This branch doesn't work on its own so it needs #447 to work. And I still haven't fully tested it. I'll switch back to draft mode until I get the time to validate and deep-dive

PR #467, opened 2026-05-16
Migrate all examples to schema version 14

akre54:
#### Background

The codebase uses a versioned schema system for project files. Examples were at different versions (v4-v7), requiring runtime migration on load. This PR migrates all examples to the latest schema version 7 and improves the migration infrastructure to run headlessly without UI dependencies.

#### Change List
- Migrated 14 examples from v6→v7 and 1 example from v4→v7 (all 17 examples now at v7)
- Consolidated all examples into `src/examples/` directory (removed `public/examples/`)
- Created `migration-utils.ts` with lightweight utilities (no store/UI dependencies)
- Updated `migrate-schema.ts` to use migration-utils, breaking dependency chain
- Fixed `renameHandle` to not add undefined fields during migration
- Fixed migration tests for new edge ID format (dots instead of colons)
- Added `verify-examples.mjs` script for validation
- Enhanced migration scripts with better logging and support for both directories


noodles-editor/vite.config.migrate.ts
```diff
@@ -11,7 +11,26 @@ export default defineConfig({
     outDir: './dist',
     emptyOutDir: false,
     rollupOptions: {
-      external: ['node:fs/promises', 'node:path'],
+      external: [
+        'node:fs/promises',
+        'node:path',
+        // External all UI/React dependencies to avoid bundling them
```
chrisgervang, line 17: Nice

noodles-editor/scripts/migrate-examples-simple.mjs
```diff
@@ -0,0 +1,123 @@
+import { readdir, readFile, writeFile } from 'node:fs/promises'
+import { join } from 'node:path'
+
+const EXAMPLES_DIRS = ['./src/examples']
+const TARGET_VERSION = 7
```
chrisgervang: Centralize across repo?

chrisgervang (approved):

noodles-editor/scripts/migrate-examples-simple.mjs
akre54: Claude says:

> We don't strictly need to centralize it because `NOODLES_VERSION` is already dynamically calculated from the migration files:
> 
> ```typescript
> // In migrate-schema.ts
> const migrations = import.meta.glob(['../__migrations__/*.ts', '!../__migrations__/*.test.ts'])
> export const NOODLES_VERSION = Math.max(...Object.keys(migrations).map(versionFromFilename))
> ```
> 
> So the version is **derived** from the presence of migration files (e.g., `014-remove-map-style-op.ts` → version 7 is 14 migrations from v0).
> 
> **For the standalone `.mjs` scripts**, we have a few options:
> 
> 1. **Keep hardcoded (current)** - Simple, but needs manual update when new migrations are added
> 2. **Import from migrate-schema.ts** - Single source of truth, but adds complexity (TypeScript import in .mjs file, requires build/loader)
> 3. **Duplicate the scanning logic** - Auto-updates, but duplicates code
> 
> **My recommendation**: Keep it hardcoded in the standalone scripts because:
> - These are **one-off migration scripts** that will rarely be run after this PR
> - The `migrate-project-files.ts` already uses `migrateProject()` which automatically uses `NOODLES_VERSION`
> - Adding TypeScript imports to `.mjs` files adds complexity for marginal benefit
> - If we add migration 015 in the future, we'd need to re-run migrations anyway and would update the version then
> 
> However, if you prefer to centralize it, I could:
> - Convert the `.mjs` scripts to `.ts` and add them to the build pipeline
> - Or export a `NOODLES_VERSION.json` file that both systems read
> 
> What do you think? Is the current approach acceptable, or would you prefer one of the alternatives?


I'm sorta leaning towards not committing the migrators to the repo, since realistically we only really care about keeping the examples up-to-date. Maybe in some future state it would make sense to combine the browser- and node-based migrations

chrisgervang: Yeah.. I'd either not commit these or bring in the dynamic import, since we've already invested in a good solution there

chrisgervang: We would be more likely to keep examples up to date if we committed migration scripts like this

PR #474, opened 2026-05-20
fix: execute isolated operators in graph executor (#471)

### LINKED ISSUE #471: [Bug] hover-previewing a fileop should display the file contents
### Description

Loading a file and hovering on it's output doesn't display the file contents:

Image

The FileOp only parses the file once it's connected to a output op, like a viewerop:

Image

### Expected Behavior

Image

File contents should load when a hover preview is performed

### Steps to Reproduce

New project
Add fileop
Load a file
Hover over output port

### Environment

- Noodles version:
- Browser:
- OS:

### Logs

_No response_

chrisgervang:
## Summary

- Removes the `upstream.size > 0` guard in `findRootOperators()` so isolated operators (no downstream connections) are treated as roots and pulled by the graph executor
- Operators like FileOp now execute as soon as their inputs are configured, regardless of whether they're connected to anything

## Root Cause

In `findRootOperators()`, operators with no downstream dependents were only treated as roots if they had upstream dependencies (`upstream.size > 0`). A standalone FileOp with no connections in either direction was excluded, so it never got pulled by the executor and its output fields retained their default values.

## What happens now (e.g. a large file loaded by an isolated FileOp)

1. **First frame after FileOp is added/configured**: The operator starts as `DIRTY`, so `pull()` executes it — fetching and parsing the file. This happens **once**.

2. **Every subsequent frame**: `pull()` is called on the operator but immediately returns the cached result because status is `CLEAN` (just an enum comparison and return — negligible cost).

3. **When a field changes** (e.g. user changes the URL): `markDirty()` sets status back to `DIRTY`. Next frame, `pull()` re-executes. Again, only once.

The per-frame cost of including isolated operators as roots is essentially zero when they're clean. The only behavioral change is that operators execute eagerly when first configured rather than waiting for a downstream connection.

## Performance 

Some projects may experience longer loading times if they have many expensive disconnected nodes that were previously not executed, but I think this is pretty reasonable behavior - if the node is a part of the project, the user probably wanted to load it.

## Test plan

- [ ] Add a FileOp to a blank project
- [ ] Load a file (CSV, JSON, or GeoJSON)
- [ ] Verify the operator executes (execution state shows success) without needing to connect it
- [ ] Hover over the output port — file contents should appear in the preview tooltip
- [ ] Connect to a ViewerOp and verify everything still works correctly
- [ ] Verify no performance regression with multiple disconnected operators in a graph

Fixes #471


akre54 (commented):
I'm not sure this is the root cause... It's that the pull status is determined by a connection to a downstream output operator (OutOp, ViewerOp, etc). Notice that connecting the FileOp to a ViewerOp will cause it to run the pull.

We may want to drop that optimization (or automatically pull all ops when they're added to the graph)

chrisgervang:
> I'm not sure this is the root cause... It's that the pull status is determined by a connection to a downstream output operator (OutOp, ViewerOp, etc). Notice that connecting the FileOp to a ViewerOp will cause it to run the pull.
> We may want to drop that optimization (or automatically pull all ops when they're added to the graph)

(Claude's analysis — I'm not familiar enough with this part of the engine to weigh in directly)

Looking at `findRootOperators()` in `graph-executor.ts:575-600`, it already includes operators with no downstream dependents as roots — but only if they have upstream dependencies (`upstream.size > 0` at line 593). A standalone FileOp with no connections at all gets excluded because it has zero upstream *and* zero downstream.

So the gap is specifically: isolated leaf nodes (no connections in either direction) never get pulled by the graph executor.

Two options the reviewer raised and their tradeoffs:

**Option A: Remove the `upstream.size > 0` guard** — Would cause every isolated operator to execute on every frame. That could regress performance for graphs with many disconnected nodes during editing (e.g. operators being assembled before wiring).

**Option B: Auto-pull when added to graph** — Cleaner, but would need to handle the case where an operator is added before its inputs are configured (e.g. FileOp added but no file selected yet).

**Current approach (hover-triggered pull):** Demand-driven — only executes when the user actively inspects. Avoids wasting cycles on operators nobody is looking at. The race condition from Greptile's review is already handled in the follow-up commit (bail out if `hoverTimerRef` was cleared during the await).

----

My thoughts: We could pivot to either alternative if you think the hover-pull is too much of a band-aid, but I think it's following the spirit of pull-base execution. The tradeoff is basically "always pay the cost" vs "pay on demand."

chrisgervang:
@greptile-apps please re-review.

@akre54 I've removed the isolated operators optimization so they are now pulled by the graph executor when dirty.

akre54:
Claude says:

## Core Problem with the Approach

**The PR solves the problem at the wrong architectural level.** It changes the fundamental execution semantics to fix a UI interaction issue.

### What Changed
- **Before**: Isolated operators (no downstream connections + has upstream inputs) did NOT auto-execute in the main loop
- **After**: ALL operators with no downstream connections now auto-execute every frame if dirty

### Why This Is Problematic

**1. Performance Impact**
- A FileOp reading a 10MB file now executes every frame instead of on-demand
- Any disconnected operator with inputs (during editing, debugging, or exploration) auto-executes
- This is frame-rate bound (60fps) vs user interaction-rate bound (hover after 1s)

**2. Semantic Shift**
- Changes the execution model from "explicit sinks pull dependencies" to "anything not consumed is also a sink"
- An operator that gets disconnected suddenly starts auto-executing when it didn't before
- Creates ambiguity: is an isolated operator "work in progress" or "should execute"?

**3. Unintended Side Effects**
- CodeOp/DuckDbOp/NetworkOp with side effects now fire automatically when disconnected
- Users can't create "dormant" operators without downstream effects
- No way to distinguish "intentionally isolated" from "temporarily isolated"

**4. Architectural Inconsistency**
The original fix attempt (commit af73e3ad) was actually architecturally cleaner: it made hover explicitly call `op.pull()`. The PR reverted this in favor of relying on the executor's root-finding logic, which was never designed for this use case.

## Better Approaches to Consider

1. **Keep the hover-driven pull** - The original approach where hover explicitly calls `await op.pull()` on the hovered operator is more semantically correct. The issue is likely in execution timing, not the approach itself.

2. **Add an "evaluate on demand" flag** - Mark operators that should execute on hover vs auto-execute

3. **UI-driven execution context** - Have a separate execution path for "preview" that doesn't alter the main graph execution semantics

## The Real Question

**Should isolated operators (with inputs but no outputs connected) automatically execute?**

The PR says "yes" but provides no rationale for why this is the correct semantic behavior beyond fixing the hover bug. This is a fundamental execution model decision being made implicitly.

## Verdict

**This is potentially problematic to merge** without deeper architectural discussion. It fixes a narrow UI bug by changing broad execution semantics, which could create performance issues and confusing behavior for users who disconnect operators while editing.

chrisgervang:
At the end of the day I'm open to a more surgical approach of only pulling ops when the user needs them. Lazy execution is a good thing, so if we can preserve that I'd favor that and it was the original approach I proposed. To me a user needs them when they want the output. I think the hover preview is a very user friendly feature, so I'm advocating we keep it.

For the implementation, the original hover-pull approach worked and was more targeted. Open to whatever fits the framework cleanest.

@akre54 this analysis contradicts what you suggested earlier, and doesn’t clarify what you’d actually agree with. What’s your position?

From a user’s perspective, if I configure a FileOp and load a file, I expect the data to be ready when I inspect it. I’m happy to revert to the hover-pull approach, but for the record the concerns about the current implementation aren’t quite right:

> This is frame-rate bound (60fps) vs user interaction-rate bound (hover after 1s)

Execution is bound by clean/dirty status. Clean ops return cached results immediately

> 3. Unintended Side Effects

Aren’t side effects discouraged? If an op has side effects, that’s a bug in the op, not a reason to prevent execution.

PR #475, opened 2026-05-21
fix: skip connected fields in property history to prevent stringify overflow

chrisgervang:
#### Background

`captureOperatorInputs()` (introduced in #362) serializes all operator field values on every focus event for undo/redo history. In projects with large datasets (e.g. CSV files with thousands of rows), connected input fields hold the full upstream data array. Attempting to `JSON.stringify` all of this caused either a `RangeError: Invalid string length` crash or a 10+ second UI freeze.

#### Change List
- Skip fields with active subscriptions (connected inputs) in `captureOperatorInputs()` — these receive values from upstream operators and don't need to be in the property history snapshot
- Add try/catch fallback around `JSON.stringify` so the app never crashes even if the state is unexpectedly large
- Add tests verifying connected fields are excluded, unconnected fields are still captured, and performance stays under 50ms even with 50 operators and large mock datasets

### DIFF
--- a/noodles-editor/src/noodles/utils/property-history.test.ts
+++ b/noodles-editor/src/noodles/utils/property-history.test.ts
@@ -291,6 +291,111 @@ describe('registerPropertyMutationCallback and firePropertyMutation', () => {
   })
 })
 
+describe('captureOperatorInputs skips connected fields', () => {
+  it('excludes fields that have active subscriptions (connected inputs)', () => {
+    const connectedField = {
+      serialize: vi.fn(() => Array.from({ length: 50000 }, (_, i) => ({ id: i }))),
+      setValue: vi.fn(),
+      subscriptions: new Map([['edge-1', {}]]),
+    }
+    const localField = mockField(20)
+    const op = mockOp('/scatter', { data: connectedField, getRadius: localField })
+    mockedGetAllOps.mockReturnValue([op as never])
+
+    const result = captureOperatorInputs()
+    const parsed = JSON.parse(result)
+
+    expect(parsed['/scatter'].getRadius).toBe(20)
+    expect(parsed['/scatter'].data).toBeUndefined()
+    expect(connectedField.serialize).not.toHaveBeenCalled()
+  })
+
+  it('includes fields with empty subscriptions (no connections)', () => {
+    const field = {
+      serialize: vi.fn(() => 'test-value'),
+      setValue: vi.fn(),
+      subscriptions: new Map(),
+    }
+    const op = mockOp('/op', { name: field })
+    mockedGetAllOps.mockReturnValue([op as never])
+
+    const result = captureOperatorInputs()
+    const parsed = JSON.parse(result)
+
+    expect(parsed['/op'].name).toBe('test-value')
+  })
+
+  it('includes fields without subscriptions property (legacy/interface fields)', () => {
+    const field = mockField('hello')
+    const op = mockOp('/op', { text: field })
+    mockedGetAllOps.mockReturnValue([op as never])
+
+    const result = captureOperatorInputs()
+    const parsed = JSON.parse(result)
+
+    expect(parsed['/op'].text).toBe('hello')
+  })
+})
+
+describe('captureOperatorInputs performance', () => {
+  it('completes in under 50ms even with many operators and large connected data', () => {
+    const ops = Array.from({ length: 50 }, (_, i) => {
+      const connectedDataField = {
+        serialize: vi.fn(() => Array.from({ length: 100000 }, () => ({ x: 1, y: 2 }))),
+        setValue: vi.fn(),
+        subscriptions: new Map([['edge', {}]]),
+      }
+      const localField = mockField(i)
+      return mockOp(`/op-${i}`, { data: connectedDataField, value: localField })
+    })
+    mockedGetAllOps.mockReturnValue(ops as never[])
+
+    const start = performance.now()
+    const result = captureOperatorInputs()
+    const elapsed = performance.now() - start
+
+    expect(elapsed).toBeLessThan(50)
+    const parsed = JSON.parse(result)
+    expect(Object.keys(parsed)).toHaveLength(50)
+    expect(parsed['/op-0'].value).toBe(0)
+    expect(parsed['/op-0'].data).toBeUndefined()
+  })
+
+  it('does not call serialize() on connected fields', () => {
+    const expensiveSerialize = vi.fn(() => {
+      throw new Error('Should not be called')
+    })
+    const connectedField = {
+      serialize: expensiveSerialize,
+      setValue: vi.fn(),
+      subscriptions: new Map([['edge-1', {}]]),
+    }
+    const op = mockOp('/op', { bigData: connectedField, radius: mockField(5) })
+    mockedGetAllOps.mockReturnValue([op as never])
+
+    const result = captureOperatorInputs()
+    const parsed = JSON.parse(result)
+
+    expect(expensiveSerialize).not.toHaveBeenCalled()
+    expect(parsed['/op'].radius).toBe(5)
+  })
+
+  it('returns empty JSON object when stringify would throw', () => {
+    const circular: any = {}
+    circular.self = circular
+    const field = {
+      serialize: vi.fn(() => circular),
+      setValue: vi.fn(),
+      subscriptions: new Map(),
+    }
+    const op = mockOp('/op', { bad: field })
+    mockedGetAllOps.mockReturnValue([op as never])
+
+    const result = captureOperatorInputs()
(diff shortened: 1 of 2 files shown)


noodles-editor/src/noodles/utils/property-history.ts
```diff
@@ -15,18 +15,25 @@ export function registerPropertyMutationCallback(cb: PropertyMutationCallback |
...
     for (const [name, field] of Object.entries(op.inputs as Record<string, IField>)) {
+      if ('subscriptions' in field && (field as { subscriptions: Map<unknown, unknown> }).subscriptions.size > 0) continue
       inputs[name] = field.serialize()
     }
     state[op.id] = inputs
   }
   debugHistorySnapshot('Captured operator inputs for %d ops', ops.length)
-  return JSON.stringify(state)
+  try {
+    return JSON.stringify(state)
```
akre54, line 32: This still is a potential performance regression, right?

noodles-editor/src/noodles/utils/property-history.ts
```diff
@@ -15,18 +15,25 @@ export function registerPropertyMutationCallback(cb: PropertyMutationCallback |
...
+  } catch {
+    debugHistory('State too large to serialize for property history')
```
akre54: Not necessarily too large. Could be for a number of reasons

noodles-editor/src/noodles/utils/property-history.ts
```diff
@@ -15,18 +15,25 @@ export function registerPropertyMutationCallback(cb: PropertyMutationCallback |
...
+    return '{}'
+  }
```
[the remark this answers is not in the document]
akre54: This is a valid review

noodles-editor/src/noodles/utils/property-history.ts
```diff
@@ -15,18 +15,25 @@ export function registerPropertyMutationCallback(cb: PropertyMutationCallback |
...
 // This mirrors captureTimelineState() in timeline-store.ts.
-export function captureOperatorInputs(): string {
+// Skips connected fields (they receive values from upstream and can hold large datasets).
+export function captureOperatorInputs(): string | null {
   const ops = getAllOps()
   const state: Record<string, Record<string, unknown>> = {}
   for (const op of ops) {
     const inputs: Record<string, unknown> = {}
     for (const [name, field] of Object.entries(op.inputs as Record<string, IField>)) {
+      if ('subscriptions' in field && (field as { subscriptions: Map<unknown, unknown> }).subscriptions.size > 0) continue
```
akre54, line 25: Doesn't the `IField` type cover the need for this conversion?

noodles-editor/src/noodles/utils/property-history.ts
```diff
@@ -15,18 +15,25 @@ export function registerPropertyMutationCallback(cb: PropertyMutationCallback |
 
 // Serializes all operator field values using each field's serialize() method.
 // This mirrors captureTimelineState() in timeline-store.ts.
-export function captureOperatorInputs(): string {
+// Skips connected fields (they receive values from upstream and can hold large datasets).
+export function captureOperatorInputs(): string | null {
```
akre54, line 19: Could we have a different value than `null`? What goes into history in that case?

Also do we even need to serialize to JSON? We should be able to just store the objects in the history array no? We can then use a shallow equals to determine equality

noodles-editor/src/noodles/utils/property-history.test.ts
```diff
@@ -291,6 +291,141 @@ describe('registerPropertyMutationCallback and firePropertyMutation', () => {
...
+
+    expect(parsed['/op'].text).toBe('hello')
+  })
+})
+
+describe('captureOperatorInputs performance', () => {
+  it('completes in under 50ms even with many operators and large connected data', () => {
+    const ops = Array.from({ length: 50 }, (_, i) => {
+      const connectedDataField = {
+        serialize: vi.fn(() => Array.from({ length: 100000 }, () => ({ x: 1, y: 2 }))),
```
akre54, line 344: Is this a valid test of the 75mb file you were using? Does this fail without the patch, and pass with it?

akre54 (approved):
Thanks for finding this! Looks like a good fix but needs some tightening. Approving to unblock

noodles-editor/src/noodles/utils/property-history.test.ts
chrisgervang: This does fail without the patch and passes with it

akre54: Awesome. Shipit

noodles-editor/src/noodles/utils/property-history.ts
chrisgervang: `null` means the history entry is skipped - the user’s change still applies but that specific edit won’t appear in undo/redo. This should only happen as a fallback when an error is thrown, which shouldn't happen normally - seems like a reasonable signal.

> Also do we even need to serialize to JSON? We should be able to just store the objects in the history array no? We can then use a shallow equals to determine equality

Maybe, but lets look into it in a follow up to think through the implications. Might be able to get rid of the null in that case as well.

noodles-editor/src/noodles/utils/property-history.ts
chrisgervang: `IField` doesn't have `subscriptions`. Should we add it? The `in` guard is kinda a middle ground if we want to keep that interface tight

noodles-editor/src/noodles/utils/property-history.ts
chrisgervang: Asked claude to look into this and there are valid points on both sides. 

> Let me trace through the implications.
> 
> **Current flow:**
> 1. Focus → `captureOperatorInputs()` → builds object → `JSON.stringify()` → stores string
> 2. Blur → `captureOperatorInputs()` again → `JSON.stringify()` → `before === after` (string compare) → if different, stores both strings in history
> 3. Undo → `JSON.parse(snapshot)` → iterates and calls `field.setValue()`
> 
> **What storing objects would look like:**
> 1. Focus → build state object → store it directly
> 2. Blur → build new state object → compare → if different, store both
> 3. Undo → iterate object and call `field.setValue()` (no parse needed)
> 
> **Implications to think through:**
> 
> 1. **Equality** — Can't use `===` on objects. Would need a two-level shallow compare: iterate ops, then fields, compare values with `Object.is`. For primitives (numbers, booleans, short strings) this is fine. But some fields serialize to objects/arrays (e.g., CodeField returns `string[]`, ColorField could return an array). Shallow equality wouldn't catch mutations inside those.
> 
> 2. **Mutability** — This is the big risk. JSON roundtrip gives you a free deep copy. If we store the raw object returned by `field.serialize()` and that object is later mutated in place, the "before" snapshot in history would be corrupted silently. We'd need to verify that every field type's `serialize()` returns a fresh value (not a reference to internal mutable state).
> 
> 3. **Memory** — Strings are immutable and can be GC'd as a unit. Object snapshots hold references to individual values, which may keep other objects alive longer.
> 
> 4. **The actual perf win** — After our fix (skipping connected fields), the remaining fields are all small primitives/short strings. The stringify cost for those is negligible. The performance problem was serializing 100K-row arrays, which we've eliminated.
> 
> **My recommendation:** This is a valid optimization but low-urgency now that connected fields are skipped. It requires auditing `serialize()` return values across all field types for mutability guarantees, and the perf gain on the remaining small fields is marginal. Better as a follow-up if profiling shows stringify is still a bottleneck.

I'm concerned about mutations on shallow copies over time. Seems safest to deep copy for a history

noodles-editor/src/noodles/utils/property-history.ts
akre54: I guess the type really should be `Field` rather than `IField`. `IField` is just there as a way to enable reflection (and all those types could probably use a refresh tbh)

noodles-editor/src/noodles/utils/property-history.ts
akre54: We don't really mutate anywhere, and I'd guess some comprehensive tests would cover these cases. In general I don't like `JSON.stringify` as a substitute for deep-equality, and it's really only because something like Immutable.js doesn't have as nice ergonomics as js objects that we don't use it. It might be worthwhile to look into functional data structures for this

chrisgervang: It's not just about equality though since it's also snapshotting values in history to recall later. I don't like stringify either, but I can see why it was used on the first turn as a catch-all

We can always try it and audit the fields. I'm having a hard time thinking of a field editable by a user that would be problematic.. plenty of field outputs are JS objects, but editable inputs are simple types

PR #476, opened 2026-05-21
feat: show executing indicator for ops taking longer than 200ms

chrisgervang:
#### Background

Operators that take a long time to execute had no visual feedback during execution in pull-based mode. The existing spinner indicator and blue border glow were wired up but never triggered.

#### Change List
- Add a 200ms delayed timer in `_pullExecution()` that sets execution state to `'executing'`, activating the existing spinner and border glow
- Clear the timer on both success and error paths so fast ops (<200ms) never flash the indicator

### DIFF
--- a/noodles-editor/src/noodles/operators.ts
+++ b/noodles-editor/src/noodles/operators.ts
@@ -496,6 +496,10 @@ export abstract class Operator<OP extends IOperator> {
 
     let executionTime = 0
     let startTime = 0
+    // Show executing indicator only for ops taking longer than 200ms
+    const executingTimer = setTimeout(() => {
+      this.executionState.next({ status: 'executing' })
+    }, 200)
 
     try {
       // Pull upstream dependencies first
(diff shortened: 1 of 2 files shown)


noodles-editor/src/noodles/operators.ts
```diff
@@ -496,6 +496,10 @@ export abstract class Operator<OP extends IOperator> {
 
     let executionTime = 0
     let startTime = 0
+    // Show executing indicator only for ops taking longer than 200ms
+    const executingTimer = setTimeout(() => {
+      this.executionState.next({ status: 'executing' })
+    }, 200)
```
akre54, line 502: What happens if the op takes less than 200ms to finish? Won't this be a race condition?

akre54 (approved):
Oh! Maybe this changed recently. Can you add a test for this?

noodles-editor/src/noodles/operators.ts
chrisgervang: `clearTimeout(executingTimer)` cancels the timer before it ever fires in that case

akre54: valid thanks

PR #479, opened 2026-05-22
feat: add GeoJSON import with table and file modes

chrisgervang:
## Summary

Add GeoJSON import support to the data importer with two modes:
- **Import as table** — flattens features into a `TableEditorOp` with properties as columns and a geometry column (point2d for Points, JSON string for others), connected to a `ViewerOp`
- **Load as file** — imports via `FileOp` connected to `GeoJsonLayerOp` → existing `DeckRendererOp`

The importer auto-detects GeoJSON content (by `.geojson` extension or parsing `.json` files) and shows a mode selection UI.

## Changes
- Extract pure import logic into `geojson-import-nodes.ts` (avoids browser globals in tests)
- Add `createGeoJsonTableDropNodes` — flattens GeoJSON to table rows with schema inference
- Add `createGeoJsonFileDropNodes` — creates FileOp + GeoJsonLayerOp pipeline
- Add `isGeoJson` type guard and `flattenGeoJsonToTable` with column type inference
- Update data importer UI with table/file radio toggle for GeoJSON files
- Add 23 tests covering both import modes, schema inference, and edge cases

## Test plan
- [ ] Import a `.geojson` file — verify table mode creates TableEditorOp + ViewerOp with correct data
- [ ] Import a `.geojson` file in file mode — verify FileOp + GeoJsonLayerOp are created
- [ ] Import a `.json` file containing a FeatureCollection — verify GeoJSON detection works
- [ ] Import a regular `.json` or `.csv` file — verify it falls through to the standard import path
- [ ] Verify Point-only GeoJSON gets `point2d` geometry column type
- [ ] Verify mixed-geometry GeoJSON gets `string` geometry column type


noodles-editor/src/noodles/operators.ts
```diff
@@ -6854,6 +6854,136 @@ export class PointOp extends Operator<PointOp> {
...
+  createOutputs() {
+    return {
+      feature: new GeoJsonField(),
+    }
+  }
+  execute({
+    geometry,
+    properties,
+  }: ExtractProps<typeof this.inputs>): ExtractProps<typeof this.outputs> {
+    const coordinates = JSON.parse(geometry)
```
akre54: I feel like the parsing should happen elsewhere

noodles-editor/src/noodles/operators.ts
```diff
@@ -6854,6 +6854,136 @@ export class PointOp extends Operator<PointOp> {
...
+    const props = JSON.parse(properties)
+    const feature = turf.multiPoint(coordinates, props)
```
akre54: Are these ops just simple wrappers to turf functions? Seems like noodles has quite a few of these

noodles-editor/src/noodles/operators.ts
```diff
@@ -6854,6 +6854,136 @@ export class PointOp extends Operator<PointOp> {
...
+  createOutputs() {
+    return {
+      feature: new GeoJsonField(),
+    }
+  }
+  execute({
+    geometry,
+    properties,
+  }: ExtractProps<typeof this.inputs>): ExtractProps<typeof this.outputs> {
+    const coordinates = JSON.parse(geometry)
```
akre54: Why is it parsing it as a string?

noodles-editor/src/noodles/operators.ts
```diff
@@ -6865,6 +6865,136 @@ export class PointOp extends Operator<PointOp> {
...
+  }
+}
+
+export class MultiPolygonOp extends Operator<MultiPolygonOp> {
+  static displayName = 'MultiPolygon'
+  static description = 'Create a GeoJSON MultiPolygon feature from coordinates'
+  asDownload = () => this.outputData
+  createInputs() {
+    return {
+      geometry: new CodeField('[]', { language: 'json' }),
```
akre54: This feels like a less-nice UI than PointOp. I know it's just a start but might be good to think about

noodles-editor/src/noodles/components/tools/data-importer-tool.tsx
```diff
@@ -117,71 +118,328 @@ function createFileDropNodes(url: string, format: string, basePosition: { x: num
   return { nodes, edges }
 }
 
+export function createGeoJsonFileDropNodes(url: string, basePosition: { x: number; y: number }) {
+  const dataId = nodeId('data', '/')
+  const geojsonLayerId = nodeId('geojson-layer', '/')
+
+  const existingDeck = getOpEntries().find(
```
akre54: Thoughts about a migration util for existing nodes / projects?

noodles-editor/src/noodles/components/tools/geojson-importer.test.ts
```diff
@@ -0,0 +1,376 @@
...
+              [-74.006, 40.7128],
+            ],
+          ],
+        },
+        properties: { name: 'Manhattan area' },
+      },
+    ],
+  }
+
+  describe('createGeoJsonDropNodes', () => {
```
akre54: Worth adding some UI and render tests. Also guarding on error cases

akre54 (commented):
Deserialization can be really expensive so I'd assume this is the kind of thing that should happen on updates, not on every execute. Probably it makes more sense to have a `ToGeoJSONOp` or similar, that has a dropdown on types (say it can switch between linestring, polygon, multipolygon, etc).

I feel like this isn't the right direction - GeoJSON is meant to house collections of features, and splitting them into individual nodes seems a bit strange to me. We should be working on datasets and not individual operators. If you have 5 points you should probably create them by hand. If you have 100 points, you should manipulate them as a collection

I'm just not really sure what the use case is... Can you expound on it a bit more? Maybe for visibility of individual points we could use a filter? Or a operator that dynamically creates checkboxes for the input toggles?

It also splits the way that you work with data into FileOps and this import method. Maybe if the goal is to be able to visualize individual geojson, we could have a `SelectFeatureOp`? For editing it could be a `EditPropertiesOp`? (TransformGeometryOp, SplitByPropertyOp, etc)

I'd also suggest looking at how TD does [Replication](https://docs.derivative.ca/Replicator_COMP). Might be a good pattern here for what you're trying to do (I think)

chrisgervang:
@akre54 The CUJ I’m trying to support is quickly importing and customizing a small geometry set that I’ve been given, which is a frequent occurrence. I agree the replicator pattern makes sense when you have a hundred of something, but not 5. Making 5–15 geometries by hand is time consuming and tedious - if I already have a GeoJSON and noodles supports geometry primitives, I’d like a fast way to get to those nodes.

Applying individual customizations to features inside a FileOp is quite difficult since you end up maintaining a chain of deltas and transforms. It’s easier to store each geometry feature as a native op and mutate it directly.

That said, I hear your point about working with datasets rather than individual ops. I’ve scoped this PR down to just a table/file import (`TableEditorOp` and `FileOp`) and am splitting the geometry feature ops into their own branch. The `SelectFeatureOp` / `EditPropertiesOp` ideas are interesting - maybe worth exploring as alternatives to the deconstruct approach.

My opinion is if it saves time for a real workflow, we should offer it. But I’m open to iterating on the right UX for the individual-feature case separately.

chrisgervang:
@greptile-apps re-review

chrisgervang:
@greptile-apps rereview

PR #481, opened 2026-05-31
fix: compute ReferenceEdges at graph transform time to fix op() resolution on load

akre54:
#### Background

On project load, `op("/path")` references in CodeOps resolved to default/empty values because ReferenceEdges (which establish execution dependencies) were only created by UI components after mounting — but the executor starts immediately. This race condition meant CodeOps executed before their referenced operators had output values.

#### Change List
- Add `computeReferenceEdges` helper in `transform-graph.ts` that parses CodeField/ExpressionField values with `getFieldReferences()` at deserialization time
- `transformGraph` now returns `{ operators, referenceEdges }` and merges synthetic ReferenceEdges into the dependency graph before the executor starts
- Update `noodles.tsx` load path to include computed ReferenceEdges in ReactFlow edge state
- Add `getUpstreamDependencies()` / `getDownstreamDependents()` accessors to Operator class
- Add 9 new tests covering reference edge synthesis (CodeOp, DuckDbOp, ExpressionOp, relative paths, self-params, missing refs, duplicates, dependency correctness)
- Update all test files for the new `transformGraph` return type

### DIFF
--- a/noodles-editor/src/noodles/__tests__/graph-execution-e2e.test.tsx
+++ b/noodles-editor/src/noodles/__tests__/graph-execution-e2e.test.tsx
@@ -51,7 +51,7 @@ describe('Graph Execution E2E', () => {
       },
     ]
 
-    const operators = transformGraph({ nodes, edges })
+    const { operators } = transformGraph({ nodes, edges })
 
     // Verify all operators created
     expect(operators).toHaveLength(3)
@@ -174,7 +174,7 @@ describe('Graph Execution E2E', () => {
       },
     ]
 
-    const operators = transformGraph({ nodes, edges })
+    const { operators } = transformGraph({ nodes, edges })
 
     expect(operators).toHaveLength(4)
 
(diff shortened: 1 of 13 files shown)


akre54:
The real issue was fixed in #483. This feels less important (though might be nice for ensuring immediate consistency at some point in the future)

chrisgervang:
> The real issue was fixed in #483. This feels less important (though might be nice for ensuring immediate consistency at some point in the future)

I think it'd be nice to have since it's easier to reason about a pure/headless execution chain. I'd like to think the UI just an editor rather than a core part of the execution lifecycle

akre54:
Sure but I think that would need need a stronger product reason than just a nice-to-have. This change duplicates the logic and spreads it out which is not ideal. Maybe the answer is decoupling it from the component lifecycle but that's a separate question

chrisgervang:
Yeah, this implementation wasn't what I had in mind - I was thinking of decoupling the graph generation from the UI components. Right now references are limited to code fields but I think it could make sense to pull out once expressions are supported on every field.

Correct me if I'm wrong, but I think the implication of the current coupling is that reference edges won't exist until the 2nd execution. The main product reason for me would be predictability - I expect the graph to execute WYSIWYG, but in reality the first execution is missing edges and can execute out of order

This hasn't been a huge problem in practice since we gracefully handle errors

PR #485, opened 2026-06-02
Remove pencil icon and show all fields by default

akre54:
#### Background

The operator properties panel required users to click a pencil icon to enter "edit mode" before they could see hidden fields or manage field visibility. This extra step made field management less intuitive.

#### Change List
- Remove PencilIcon component and edit mode state/toggle
- Always render all input fields (both visible and hidden) in a unified list
- Show add (+) buttons for hidden fields, remove (−) buttons for visible fields at all times
- Simplify Reset button logic (no longer gated by edit mode)
- Remove "Hidden fields" divider, search functionality, and "Show all" button
- Update tests to match new always-visible behavior

### DIFF
--- a/noodles-editor/src/noodles/components/node-properties.tsx
+++ b/noodles-editor/src/noodles/components/node-properties.tsx
@@ -186,26 +186,6 @@ function Tooltip({
   )
 }
 
-function PencilIcon({ onClick, isActive }: { onClick: () => void; isActive: boolean }) {
-  return (
-    <svg
-      className={cx(s.editIcon, { [s.editIconActive]: isActive })}
-      onClick={onClick}
-      viewBox="0 0 24 24"
-      width="14"
-      height="14"
-      fill="none"
-      stroke="currentColor"
-      strokeWidth="2"
-      strokeLinecap="round"
-      strokeLinejoin="round"
-    >
-      <title>{isActive ? 'Exit edit mode' : 'Edit fields'}</title>
-      <path d="M17 3a2.85 2.85 0 1 1 4 4L7.5 20.5 2 22l1.5-5.5Z" />
-      <path d="m15 5 4 4" />
-    </svg>
-  )
-}
 
 function AddRemoveButton({
   type,
@@ -407,10 +387,8 @@ export function NodeProperties({ nodeId }: { nodeId: string }) {
   const dragDataRef = useRef<{ inputName: string; index: number } | null>(null)
   const [isDescriptionExpanded, setIsDescriptionExpanded] = useState(false)
   const [isTruncated, setIsTruncated] = useState(false)
-  const [isEditMode, setIsEditMode] = useState(false)
   const [isResetDialogOpen, setIsResetDialogOpen] = useState(false)
   const [pendingHideField, setPendingHideField] = useState<string | null>(null)
-  const [hiddenFieldSearch, setHiddenFieldSearch] = useState('')
   const [contextMenu, setContextMenu] = useState<{
     x: number
     y: number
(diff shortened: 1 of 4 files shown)


chrisgervang:
Could you share more about why you'd like to change this? I thought the idea was to keep the UI uncluttered

akre54:
For sure. Mainly discoverability of this feature, and the fact that the sidebar has a lot more space to take up than nodes do. Especially so now that the sidebar is entirely owned by Noodles. 

Fields show up both as connectable handles and as inputs in the sidebar

In Blender and Unreal Blueprints there are only input handles so the goal is to keep the number of fields on the node small. In Cavalry there are a bunch of input and output handles and all are unified on the sidebar. On Houdini and TD there are a limited number of input and output handles (usually less than 3), but potentially really big numbers of input fields in the sidebar.

Noodles sits somewhere between. We also don't yet have a way to connect arbitrary fields (#3 for starters, probably others). But we're less like Blender and more like TD or Cavalry. I think the solution to long lists of fields is to do Houdini or TD-style pages, and enable arbitrary field connections, not encouraging manual organization

chrisgervang (approved):
Yeah that all makes sense. Seem to address the discoverability issues first mentioned in #283 when this was introduced. Poking around a bit I don't mind having them always visible and it's fewer clicks, which is nice.

> Fields show up both as connectable handles and as inputs in the sidebar

This made me think of another UX improvement - what if we could drag an edge onto an input in the sidebar and it'd make the connection _plus_ make the field visible?

1. Select node B
2. Drag an edge out of node A to inputs on B (node B is still selected)
3. Edge is created from A to B

Sounds like one swift action if it doesn't make the UX clunky in some other way.

The way it is today also isn't bad, it's just one additional click

noodles-editor/src/noodles/components/node-properties.tsx
```diff
@@ -833,59 +798,8 @@ export function NodeProperties({ nodeId }: { nodeId: string }) {
...
-                            setHiddenFieldSearch('')
-                          }}
-                        >
-                          {hiddenFieldSearch ? 'Show matches' : 'Show all'}
-                        </button>
-                      </div>
-                      <input
-                        type="text"
-                        className={s.fieldSearch}
-                        placeholder="Search fields..."
```
chrisgervang, line 873: Why remove search? It helps discoverability on long property lists

akre54: Because theoretically you can do this with browser search, and hopefully none of the lists are that long. If we start having huge operators like that we should introduce pages (a la TD / Houdini) instead

akre54:
> This made me think of another UX improvement - what if we could drag an edge onto an input in the sidebar and it'd make the connection _plus_ make the field visible?

Yes definitely! TD does this with ["exporting"](https://derivative.ca/UserGuide/Export). Houdini and Blender support copying references to parameters / channels.

Ideally most Noodles programs are pretty simple but if we need to go this route absolutely can. This would tie into #3 the strongest

noodles-editor/src/noodles/components/node-properties.tsx
chrisgervang: The right way to do this is _scoped_ browser search, which still involves custom UI. Browser search that leaves the scope of the sidebar would be a poor experience - imagine having multiple hits in the graph area since words are often repeated in multiple places.

Pages make sense when there are logical groupings like `data` and `visual`, and you know what you're looking for but they're not perfect either

Search is for when you don't know where something is you're still learning the system, or you know exactly what you want.

GeoJsonLayer and the other layers are good challenges to think though.

akre54: I mean sure but realistically how many fields is it? It's not *that* much scrolling.

And if we have an organization issue lets solve it at a higher level than search

PR #491, opened 2026-06-04
feat: implement Arrow zero-copy architecture for SQL execution

akre54:
## Summary

This PR implements **Arrow zero-copy architecture** and **Houdini-style binary attributes** for Noodles.gl, delivering 2x performance improvements and a cleaner attribute-based workflow.

#### 🚀 Arrow Zero-Copy Architecture
- Keep Arrow tables from DuckDB instead of converting to JS arrays
- Zero-copy data flow: SQL → Arrow → GPU
- 2x performance improvement for large datasets
- Foundation for future WebGPU integration

#### 🎨 Houdini-Style Binary Attributes (Phase 1)
- **New**: `CreateAttributeOp` with string/boolean attribute support
- **Explicit attribute creation**: All attributes (position, color, text) created as visible nodes
- **Inspectable data flow**: See attribute values in the graph
- **Automatic migration**: AccessorOp nodes automatically converted to CreateAttributeOp at load time
- **Mixed attribute types**: Numeric attributes (GPU TypedArrays) + string/boolean attributes (CPU accessors)

### What Changed

**Before (AccessorOp pattern):**
```
Data → AccessorOp (hidden function) → Layer.getPosition
```

**After (Houdini-style):**
```
Data → CreateAttributeOp (name: 'position', expression: '[d.lng, d.lat]') → Layer (reads 'position' attribute)
```

### Migration & Compatibility

- ✅ **Automatic migration**: Projects with AccessorOp nodes automatically migrate at load time
- ✅ **Zero breaking changes**: Backward compatible with existing projects
- ✅ **78% migration rate**: Most AccessorOps convert successfully (pass-through patterns kept intentionally)
- ✅ **String attributes**: getText, getIcon now work as binary attributes

### Testing

**Comprehensive testing across:**
- 5+ complex Joby visualization projects (55-58 nodes)
- All 17 built-in examples
- String attribute migration (text labels working)
- Binary attribute rendering
- Timeline integration

**Test Results:**
- ✅ Zero critical errors
- ✅ nyc-taxis example CreateAttributeOps properly wired
- ✅ Layout algorithm simplified (400px column spacing)
- ✅ simple-mesh-example WebGL errors fixed

### Performance

- **2x faster** for datasets with 10K+ rows
- Eliminates 2-3ms Arrow→JS conversion per 10K rows
- Zero-copy column access and slicing
- GPU-ready binary attributes

### Documentation

- `.context/comprehensive-testing-summary.md` - Full testing report
- `.context/houdini-phase1-summary.md` - Houdini-style implementation details
- `.context/arrow-zero-copy-testing-report.md` - Performance analysis

### Known Edge Cases (Not Blockers)

- Some PathLayer `getPath` accessors have initialization issues (isolated, under investigation)
- External API 401 errors (pre-existing, expired API keys)

### Next Steps (Phase 2)

- Deprecate MapRangeOp → migrate to CreateAttributeOp expressions
- Deprecate ColorRampOp → use `d3.interpolate*` in CreateAttributeOp
- Add AttributeWrangleOp (VEX-style multi-attribute creation)
- Attribute inspector panel
- Attribute spreadsheet view

---

**Branch Status:** 88 commits, ready for squashing into ~8-10 logical groups


akre54:
Chatted through duckdb->geoarrow buffers for Deck with @ilyabo this morning. 

Apparently [duckdb stores all geometry](https://duckdb.org/docs/current/sql/data_types/geometry#geometry-storage) as WKB which incurs a serialization overhead to convert to GeoArrow

This PR does something similar to the [duck_geoarrow community extension](https://duckdb.org/community_extensions/extensions/duck_geoarrow), so at the moment WKB decode is unavoidable. We're caching the output so it's a one-time decode but it's not true zero-copy but zero-copy for the GPU upload once decoded.

Separately: Ilya mentioned @kylebarron is working on https://github.com/geoarrow/geoarrow-js/issues/54 which looks highly relevant. It’s a full WKB to GeoArrow parser with lazy zero-copy WKB views. Kyle, would it make sense to depend on geoarrow-js once that lands, rather than maintaining our own WKB decoder?

kylebarron:
> Apparently [duckdb stores all geometry](https://duckdb.org/docs/current/sql/data_types/geometry#geometry-storage) as WKB which incurs a serialization overhead to convert to GeoArrow

Technically duckdb does not store data as WKB, but as its own internal format. Then it has to be serialized to a valid GeoArrow array format on export from DuckDB. In theory DuckDB could choose any GeoArrow array type, but in practice the default is WKB.

> It’s a full WKB to GeoArrow parser with lazy zero-copy WKB views

The GeoArrow buffers are not zero-copy views into the WKB (they can't be, they need to be aligned and contiguous); but it's ideally only one copy from WKB into the GeoArrow buffers.

> Kyle, would it make sense to depend on geoarrow-js once that lands, rather than maintaining our own WKB decoder?

The goal is for geoarrow-js and deck.gl-geoarrow to combine to make it seamless to go from DuckDB Spatial to deck.gl, but I don't have time to work on the WKB parser right now. (Another option is to compile the [WKB parser from geoarrow-rs](https://docs.rs/geoarrow-array/latest/geoarrow_array/cast/fn.from_wkb.html) to WASM)

akre54:
Ok great thank you for that info. Super helpful, and good to know about the internals. 

> In theory DuckDB could choose any GeoArrow array type, but in practice the default is WKB.

Ok so you're saying no matter what there's an extra hop (and possibly 2 if it goes from internal duckdb -> WKB -> GeoArrow). That's a bummer but makes sense. Guessing making it contiguous wouldn't fit with the duckdb model

> The goal is for geoarrow-js and deck.gl-geoarrow to combine to make it seamless to go from DuckDB Spatial to deck.gl, but I don't have time to work on the WKB parser right now

Ok awesome. That sounds like a good vision. I imagine lots of apps could benefit from that

Zero-copy isn't a burning concern for me, more of a nice-to-have to keep the implementation straightforward and rendering fast. I think in my perfect world we'd split the execution between the CPU and GPU, and a query planner would be able to decide where to do the split

If this ever gets higher priority for either of us I'd love to talk it through

kylebarron:
> Ok so you're saying no matter what there's an extra hop (and possibly 2 if it goes from internal duckdb -> WKB -> GeoArrow). That's a bummer but makes sense. Guessing making it contiguous wouldn't fit with the duckdb model

Yes, internal duckdb geometries are not contiguous with their coordinates in the way a GPU needs. You're always going to have a copy. As long as you aren't creating too many tiny objects and slamming the GC it's fine.

> Zero-copy isn't a burning concern for me, more of a nice-to-have to keep the implementation straightforward and rendering fast. I think in my perfect world we'd split the execution between the CPU and GPU

I think you'll always need the CPU to do the transformation from a source format to a GPU-friendly format

akre54:
> I think you'll always need the CPU to do the transformation from a source format to a GPU-friendly format

Totally. But for my purposes the question is what can be cached / memoized so we don't have to copy / transform as much. Like I said even this PR is a big step up from all of the looping and copying from JS. Just trying to keep everything fast

charlieforward9:
Awesome to see noodles adopting the deck10 direction so quickly, Ricky Bobby vibes "I just wanna go fast"

noodles-editor/src/noodles/sql-compiler/PERFORMANCE-TABLE.md
charlieforward9, line 1: This would be awesome to see a little screen recorded demo within PR if not a big lift. At least current and target.

PR #498, opened 2026-06-13
feat: detect geometry column in Point2D/3DField for NetworkOp compatibility

chrisgervang:
#### Background

Screenshot 2026-06-12 at 10 10 41 PM

When connecting a TableEditorOp with a point2D column to a NetworkOp, the geometry data wasn't being detected because Point2D/3DField only looked for top-level `lng`/`lat` keys or GeoJSON Features — not a `geometry` column containing point data.

The `geometry` column name is the standard convention across the geospatial ecosystem:
- **DuckDB Spatial** — default column name when loading spatial data
- **GeoJSON (RFC 7946)** — every Feature has a `"geometry"` property
- **GeoParquet** — specifies `"geometry"` as the default column name
- **GeoPandas** — `.geometry` is the default column accessor

#### Change List
- Extend `extractGeoJsonPointCoordinates` to handle bare GeoJSON Point geometry objects (`{type: "Point", coordinates: [...]}`) in addition to Features
- Add `extractGeometryColumn` helper that detects a `geometry` property on row objects and extracts point coordinates from tuples or GeoJSON Point geometries
- Wire geometry column detection into Point3DField (both 3D and 2D arms) and Point2DField normalization
- Add 12 field-level tests for bare GeoJSON Point and geometry column extraction
- Add 8 NetworkOp integration tests covering geometry column formats and backward compatibility

### DIFF
--- a/noodles-editor/src/noodles/fields.test.ts
+++ b/noodles-editor/src/noodles/fields.test.ts
@@ -1168,6 +1168,30 @@ describe('Point2DField', () => {
     field.setValue(geoJsonFeature)
     expect(field.value).toEqual([-118.4182302, 34.0576856])
   })
+
+  it('extracts coordinates from bare GeoJSON Point geometry', () => {
+    const field = new Point2DField()
+    field.setValue({ type: 'Point', coordinates: [-73.78, 40.64] })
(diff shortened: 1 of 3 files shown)


chrisgervang:
@greptile-apps rereview

PR #500, opened 2026-06-13
fix: Point field schema not transforming tuples to objects

chrisgervang:
#### Background

Before
Screenshot 2026-06-12 at 8 49 08 PM

After
Screenshot 2026-06-12 at 8 56 23 PM

The PointOp's coordinates field displayed 0,0 even when valid coordinates were set (e.g. via Create Point). The root cause was a JS destructuring default bug: `createSchema({ returnType }: Opts = { returnType: 'object' })` only applied the default when called with no argument, but the Field constructor always passed `{}`, leaving `returnType` as `undefined` and skipping the tuple-to-object Zod transforms.

#### Change List
- Fix `Point2DField.createSchema` destructuring to use `{ returnType = 'object' }` so the default applies even when called with an empty options object
- Apply same fix to `Point3DField.createSchema`

### DIFF
--- a/noodles-editor/src/noodles/fields.test.ts
+++ b/noodles-editor/src/noodles/fields.test.ts
@@ -23,6 +23,8 @@ import {
   StringField,
   StringLiteralField,
   UnknownField,
+  Vec2Field,
+  Vec3Field,
 } from './fields'
 import { NumberOp } from './operators'
 import { clearOps, setOp } from './store'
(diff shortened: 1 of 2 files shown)


chrisgervang:
@greptile-apps please rereview

akre54 (approved):
Thanks! Can you please add some red/green tests for this?

chrisgervang:
> Thanks! Can you please add some red/green tests for this?

yup, added coverage for missing fields too

PR #503, opened 2026-06-16
fix: container scope-boundary bugs (copy-paste, delete cascade, undo/redo)

chrisgervang:
## Summary

Operations that cross container scope boundaries were silently losing children. Three bugs with the same root cause: code that queried ReactFlow's store only saw the current scope (displayed nodes), but container children live in a different scope and use path-based nesting (`/container/child`), not ReactFlow's `parentId`.

- **Copy-paste**: Copying a container dropped its children because `toObject()` only returns displayed nodes
- **Delete cascade**: Deleting a container left orphaned children in the graph
- **Undo/redo**: Undo couldn't restore children because the "before" snapshot only captured displayed nodes

## Changes

**Bug fixes:**
- Copy phase now finds container children via `getParentPath()` from the full node array
- Delete paths use `expandDeleteSet()` to cascade deletion to all path-based descendants
- Undo/redo captures full graph state (including out-of-scope children) via `graphRef`

**Refactors:**
- Unify `nodesRef`/`edgesRef` into single `graphRef` (eliminates stale-ref window from useEffect sync)
- Extract pure utilities into `copy-paste-utils.ts` (`collectContainerChildren`, `remapPastedIds`, `expandDeleteSet`, etc.)
- Add `GraphRef` type to `types.ts`

**Tests:**
- Component-level test for `CopyControls` proving it reads from `graphRef` (breaks if wiring reverted)
- Hook-level tests for `useUndoRedo` verifying `graphRef` captures full state across scopes
- Unit tests for `expandDeleteSet` and orphan invariants in `copy-paste-utils.test.ts`
- Integration test for `transformGraph` container cascade cleanup

## Test plan

- [x] Unit tests for all pure utilities pass (expandDeleteSet, collectContainerChildren, remapPastedIds)
- [x] Component test: CopyControls includes children from graphRef in clipboard
- [x] Hook test: useUndoRedo nodesBefore contains all 5 nodes (not just 3 displayed)
- [x] Hook test: undo diff correctly identifies container + children for restoration
- [x] Integration test: transformGraph cleans up operator store after cascade delete
- [x] Invariant tests: no orphaned children after delete or paste operations
- [ ] Manual: copy a container at root scope, paste, verify children are present
- [ ] Manual: delete a container, undo, verify children restored
- [ ] Manual: delete a nested container, verify no orphans remain


noodles-editor/src/noodles/__tests__/node-operations-integration.test.tsx
```diff
@@ -88,18 +91,28 @@ function createTestGraph(options: TestGraphOptions = {}) {
...
+        id: '/container/container-output',
+        type: 'GraphOutputOp',
+        position: { x: 0, y: 0 },
+        data: { inputs: {} },
+      },
       {
         id: '/container/child1',
         type: 'MathOp',
         position: { x: 50, y: 50 },
-        parentId: '/container',
```
chrisgervang, line 95: The real `createNodesForType('ContainerOp', ...)` in `node-creation-utils.ts` never sets `parentId`. Instead, it relies on the path prefix (`/container/child1`) for containment. 

This test was setting `parentId` manually, which made the test’s `copyPasteNodes` helper (which checked `node.parentId`) work even though the production code using the same `parentId` check was always getting zero children.

Re-implementing production logic in test helpers is not helping us - it creates a false sense of coverage that can mask bugs indefinitely. While copy-paste is primarily a UI integration, the complexity of its ID remapping logic pure utility functions that can be tested directly without re-implementation.

noodles-editor/src/noodles/noodles.tsx
```diff
@@ -1517,7 +1517,11 @@ export function getNoodles(): Visualization {
               <Background />
               <Controls position="bottom-right" />
               <BlockLibrary ref={blockLibraryRef} reactFlowRef={reactFlowRef} />
-              <CopyControls ref={copyControlsRef} />
+              <CopyControls
+                ref={copyControlsRef}
+                getAllNodes={() => nodesRef.current}
```
akre54: This feels like a strange pattern to me. This should either forward the React reference or make the copy controls container-aware

akre54 (approved):
Good find! If it works and tests are comprehensive merge away

noodles-editor/src/noodles/noodles.tsx
chrisgervang: We need synchronous access to the full graph detached from rendering. A store/context would work but is a larger refactor. A forwarded ref is technically possible, but it would mean creating a new ref object just to hold `{ nodes, edges }` and passing it down, just with `.current.nodes` instead of `getAllNodes()`.

chrisgervang:
@greptile-apps re-review

chrisgervang:
@greptile-apps re-review

noodles-editor/src/noodles/noodles.tsx
chrisgervang: Switched to refs and got rid of a top level effect to avoid slightly stale node and edge refs to avoid race conditions

noodles-editor/src/noodles/__tests__/node-operations-integration.test.tsx
akre54: Yeah I think parentId was a remnant of an old format (and I think react-flow has parentNode which we use for contained nodes like For Loop). Definitely good to fix if it's sticking around

PR #504, opened 2026-06-16
feat: improve properties panel context menu and visual alignment

chrisgervang:
## Summary

- Enrich the right-click context menu with Copy value/name/code reference/mustache reference actions, Hide/Show field toggle, and Disconnect all inputs
- Show all menu items always (disabled when not applicable) instead of conditionally hiding them
- Combine Sequence / Make static into a single toggle based on whether the field has keyframes
- Polish vertical alignment: all sidebar sections share a consistent 6px grid (header, description, section titles, property rows, dividers)
- Add Properties Panel documentation page with screenshots to the website

## Details

**Context menu improvements:**
- "Copy value" disabled for non-value fields (data, layers, etc.)
- "Reset to default" disabled when value equals default or field is connected
- "Hide field" disabled when field has an incoming connection
- "Disconnect all inputs" enabled only for ListFields with connections
- Auto-flip menu leftward when it would overflow the viewport

image

**Visual polish:**
- Property rows have hover highlight and 6px horizontal padding
- Output dots align with input dots (24px left padding for missing +/- button)
- Header divider, capsule pill, Reset button, and keyframe diamonds all share consistent vertical grid columns
- +/- buttons and Reset button are non-user-selectable
- Panel width set via CSS variable (296px)

**Documentation:**
- New `docs/users/properties-panel.md` with overview, context menu reference table, and screenshots
- New `dev-docs/screenshot-guide.md` with Chrome DevTools MCP workflow for capturing/cropping screenshots with fake cursor

**Tests:**
- 16 new unit tests covering context menu actions (copy, sequence, hide/show, disconnect)

## Screenshots

image

## Test plan

- [x] `npm test` passes (40 tests + 16 new, 4 pre-existing skips)
- [x] Right-click any property row → context menu appears with all actions
- [x] Copy actions write correct values to clipboard
- [x] Hide/Show field works via context menu and +/- buttons
- [x] Disabled states: connected fields can't be hidden or reset, non-value fields can't copy value
- [x] Visual alignment consistent across header, description, section titles, and property rows
- [x] Website docs build with new properties-panel page and screenshots


akre54 (changes requested):
It wasn't accidentally removed. It's now available as a right-click, which aligns with other software ([Blender Drivers](https://docs.blender.org/manual/en/latest/animation/drivers/usage.html#copy-paste), [TD export vs paste relative reference](https://matthewragan.com/2014/06/01/understanding-referencing-touchdesigner/), [Houdini parameter references](https://www.sidefx.com/docs/houdini/network/copying.html), etc). It also keeps the visual noise down.

Do you feel strongly the other way?

chrisgervang:
> It wasn't accidentally removed. It's now available as a right-click, which aligns with other software ([Blender Drivers](https://docs.blender.org/manual/en/latest/animation/drivers/usage.html#copy-paste), [TD export vs paste relative reference](https://matthewragan.com/2014/06/01/understanding-referencing-touchdesigner/), [Houdini parameter references](https://www.sidefx.com/docs/houdini/network/copying.html), etc). It also keeps the visual noise down.
> 
> Do you feel strongly the other way?

Ah ok, seemed accidental since unused `referenceIcon` CSS was left behind.

I think the menu needs to be more discoverable - the problem with the current UX is nothing on the row indicates there's a context menu.

I'll take another crack at this UX with a kebab menu. Visual noise can be kept low by making the menu only visible on hover

chrisgervang:
@greptile-apps rereview

chrisgervang:
@greptile-apps rereview

akre54 (approved):
I'm not a fan of the visual noise or the fact that it takes up the limited horizontal real estate, but it does make this more discoverable. Overall neutral on this change

Reduced horizontal real estate before:

Screenshot 2026-06-16 at 10 13 09 AM


After:
Screenshot 2026-06-16 at 10 12 18 AM

chrisgervang:
Those screenshots show a number of issues with the design, and I've pushed up some improvements:
- small inputs like checkboxes weren't shrinking, so field names were more likely to be hidden
- the sidebar became narrower overall, and I think the previous width was better
- there was extra padding added with the hover hint that was unnecessary

Also, the right padding for the invisible kebab was actually pretty problematic for alignment on the right side. Compound props complicated the layout alignment since they don't have a diamond or input on their field row.

I took a step back and decided the hover hint gives a clue that the row is interactive, and the core discoverability issues is better solved with documentation than UI.

Screenshot 2026-07-06 at 4 07 23 PM

We're basically back to where we were, but with more clues and docs. Also the context menu now includes all of the actions a user can take from the row, which doubles as another UI clue for learning the tool.

chrisgervang:
@greptile-apps rereview

PR #507, opened 2026-07-06
Agent-ready docs program: roadmap + 7 executable sub-plan specs

chrisgervang:
#### Review guide

Specs only, no implementation: a roadmap README plus seven sub-plans under `dev-docs/specs/agent-ready-docs/`. None of this is locked in; now's the cheap time to bend it.

(Updated 7/6: 03 and 05 are reconciled against #508, which landed 05's core before this was reviewed. 05 is now mostly a follow-ups spec. And #509 is drafting sub-plan 07's phase 0, stacked on this branch: the harness plus a real T0 baseline.)

**Reading order, roughly:**
1. `README.md` for the vision, CUJs, dependency graph, phasing, and the cross-cutting rules index (the program's recurring principles, each pointing at the sub-plan section that owns it). If the direction feels off here, probably not worth reading further until we talk.
2. `01-operator-reference.md` is the core of it. I think the interesting bits are the marker-based generated regions (D1), the NumberOp example page (D2, does the Chen voice actually land or is it trying too hard?), the LLM drafting protocol (D6), and the batch workflow. That last one is built around ~10 minutes of your memory per batch, and I'm curious whether the question-sheet loop sounds workable or like a chore I've dressed up.
3. For 02-05, skimming the Design sections is probably enough. They're numbered decisions with rationale, and the steps below them are execution detail.
4. `07-cuj-evals.md` is how we'd know any of this is working. The T0 baseline would run before anything lands (it has: #509 is the draft, with the first scorecard).

**Stuff I'd especially like your read on:**
- A few of these propose consolidating things you know the history of way better than I do: `generate-context.yml` (02-D3), `mcp-proxy.js` folding into `npx noodles-mcp --live` (03-D4, now likely superseded by #508's relay), unifying the tool surfaces (05, since landed via #508). I wrote down what I think the original intent was in each case, and I may well have misread some of it. If so, the spec should change, not the code.
- One deliberate reversal: the specs make the `timeline` JSON a supported editing surface (schema, lint, worked examples) instead of keeping the old don't-hand-edit rule, which looked like Theatre-era advice now that the native format is fully typed in `src/timeline/types.ts`. If there's a reason it should stay off-limits that I'm not seeing, 02/03/04/07 all need adjusting.
- Some things I treated as findings that might just be me missing context: the external-control write path looking unfinished, the two Pages publish mechanisms conflicting, the AGENTS.md version drift (6 vs 14).
- Is 7 sub-plans the right cut for two people, or would you merge some?

[...]

PR #508, opened 2026-07-06
Add WebMCP support for external control from Claude Code

akre54:
#### Background

External control of a running Noodles session previously required the WebSocket MCP proxy (`examples/external-control/mcp-proxy.js`), a separate Node process with its own hand-copied camelCase tool surface. WebMCP — the W3C `navigator.modelContext` draft — lets the page itself expose MCP tools, so clients like Claude Code connect with just `claude mcp add webmcp -- npx -y @mcp-b/webmcp-local-relay@4`. Verified end-to-end: tools listed and called over MCP stdio against a live tab, with `apply_modifications` mutating the editor graph in real time.

#### Change List
- Register the full in-app AI tool surface (~22 tools) on `navigator.modelContext` via `@mcp-b/global` under the existing `?externalControl=true` flag (lazy chunk, no main-bundle impact)
- New `src/ai-chat/tool-definitions.ts` as the single source of truth for tool schemas, consumed by both the chat panel (`claude-client.ts` refactor) and WebMCP; adds JSON Schemas for the 9 code-search/docs tools that had none
- New `src/webmcp/` module: registration with AbortSignal teardown, MCP content-block conversion (screenshots as image blocks, failures as `isError`), localhost-only relay embed injection (pinned to relay v4 — the discovery handshake changes between majors)
- Bridge from `noodles.tsx` syncs live project state and wires `useProjectModifications`, so `apply_modifications` actually mutates the live editor graph (the legacy WS path only validated)
- 18 new unit tests (schema parity, registration, error conversion, applier bridge); docs updated to recommend WebMCP with the WebSocket proxy kept as legacy

### DIFF
--- a/noodles-editor/package.json
+++ b/noodles-editor/package.json
@@ -15,6 +15,7 @@
     "@mapbox/mapbox-gl-geocoder": "^5.1.2",
     "@mapbox/polyline": "^1.2.1",
     "@math.gl/web-mercator": "^4.1.0",
+    "@mcp-b/global": "^3.0.0",
     "@microlink/react-json-view": "^1.31.15",
     "@observablehq/plot": "^0.6.17",
     "@radix-ui/colors": "^3.0.0",
@@ -60,7 +61,8 @@
     "validator": "^13.15.26",
     "web-vitals": "^5.2.0",
     "wouter": "^3.9.0",
-    "zod": "^4.3.6"
+    "zod": "^4.3.6",
+    "zod-to-json-schema": "^3.25.0"
   },
   "overrides": {
     "d3-scale-chromatic": "3.1.x"
@@ -94,6 +96,7 @@
   },
   "devDependencies": {
     "@biomejs/biome": "^2.4.9",
+    "@mcp-b/webmcp-types": "^3.0.0",
     "@testing-library/dom": "^10.4.1",
     "@testing-library/jest-dom": "^6.9.1",
     "@testing-library/react": "^16.3.2",
--- a/noodles-editor/src/ai-chat/tool-definitions.ts
+++ b/noodles-editor/src/ai-chat/tool-definitions.ts
@@ -1,0 +1,395 @@
+// Canonical tool definitions shared by the in-app Claude chat (claude-client.ts)
+// and the WebMCP registration (src/webmcp/). Single source of truth so the
+// Anthropic input_schema and the navigator.modelContext inputSchema can't drift.
+
+import type { MCPTools } from './mcp-tools'
+import type { NoodlesProject, SearchCodeParams, ToolResult } from './types'
+
+// JSON Schema subset accepted by both Anthropic tools and WebMCP registerTool
+export interface ToolInputSchema {
+  type: 'object'
+  properties: Record<string, unknown>
+  required?: string[]
+}
+
+export interface ToolDefinition {
+  // snake_case, matches the in-app chat tool names
+  name: string
+  description: string
+  inputSchema: ToolInputSchema
+  // false = executable but not offered to the in-app chat (keeps chat token budget unchanged)
+  exposeToChat?: boolean
+  // getProject supplies the live project for tools that need it injected (analyze_project)
+  execute: (
+    tools: MCPTools,
+    params: Record<string, unknown>,
+    getProject: () => NoodlesProject | null
+  ) => Promise<ToolResult> | ToolResult
+}
+
+export const toolDefinitions: ToolDefinition[] = [
+  // Visual debugging tools
+  {
+    name: 'capture_visualization',
+    description:
+      'Capture a screenshot of the current visualization. The screenshot will be attached to your next message so you can see it.',
+    inputSchema: {
+      type: 'object',
+      properties: {
+        includeUI: { type: 'boolean' },
+        format: { type: 'string', enum: ['png', 'jpeg'] },
+        quality: { type: 'number', description: 'JPEG quality 0-1, default 0.7' },
+      },
+    },
+    execute: (tools, params) =>
+      tools.captureVisualization(
+        params as { includeUI?: boolean; format?: 'png' | 'jpeg'; quality?: number }
+      ),
+  },
+  {
+    name: 'get_console_errors',
+    description: 'Get recent browser console errors and warnings',
+    inputSchema: {
+      type: 'object',
+      properties: {
+        since: { type: 'number' },
+        level: { type: 'string', enum: ['error', 'warn', 'all'] },
+        maxResults: { type: 'number' },
+      },
+    },
+    execute: (tools, params) =>
+      tools.getConsoleErrors(
+        params as { since?: number; level?: 'error' | 'warn' | 'all'; maxResults?: number }
+      ),
+  },
+  {
+    name: 'get_render_stats',
+    description: 'Get deck.gl rendering statistics',
+    inputSchema: {
+      type: 'object',
+      properties: {},
+    },
+    execute: tools => tools.getRenderStats(),
+  },
+  {
+    name: 'inspect_layer',
+    description: 'Get layer information',
+    inputSchema: {
+      type: 'object',
+      properties: {
+        layerId: { type: 'string' },
+      },
+      required: ['layerId'],
+    },
+    execute: (tools, params) => tools.inspectLayer(params as { layerId: string }),
+  },
+  // Project state tools
+  {
+    name: 'apply_modifications',
+    description:
+      'Apply modifications to the project (add/update/delete nodes or edges). Use this instead of returning JSON in text.',
+    inputSchema: {
+      type: 'object',
+      properties: {
+        modifications: {
+          type: 'array',
+          description: 'Array of modifications to apply',
+          items: {
+            type: 'object',
+            properties: {
+              type: {
+                type: 'string',
+                enum: ['add_node', 'update_node', 'delete_node', 'add_edge', 'delete_edge'],
+              },
+              data: {
+                type: 'object',
+                description: 'The node or edge data',
+              },
+            },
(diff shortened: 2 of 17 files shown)


noodles-editor/package.json
```diff
@@ -60,7 +61,8 @@
     "validator": "^13.15.26",
     "web-vitals": "^5.2.0",
     "wouter": "^3.9.0",
-    "zod": "^4.3.6"
+    "zod": "^4.3.6",
+    "zod-to-json-schema": "^3.25.0"
```
chrisgervang, line 65: zod-to-json-schema looks unused in the diff, leftover from something else?

noodles-editor/src/webmcp/register.ts
chrisgervang, line 1: Not a blocker, more curiosity as I was reading the registration path. Have you considered gating the write tools (apply_modifications, set_keyframe, delete_keyframe, set_playback_position) behind something explicit, like `?agentAccess=read-write`? Currently, registration isn't host-gated, so `externalControl=true` on the hosted app would hand writes to any browser-resident agent that happens to be installed. Maybe that's the right move for now but something to think about

Related, would skipping registration when `window.top !== window` make sense, so an embedding page can't quietly pick up the tools?

There's a fuller consent sketch in the 05 spec on the #507 branch, but honestly it might be over-designed, would be interested in your read.

noodles-editor/src/webmcp/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { initWebMCP } from './register'
+
+// The npx local relay (@mcp-b/webmcp-local-relay) only discovers tabs that load
+// its embed script; the WebMCP extension and native Chrome paths need nothing.
+// Pinned CDN build, injected on localhost only — same trust model as the
+// legacy WebSocket bridge. The pinned version must match the relay major the
+// docs tell users to run (the discovery handshake changes between majors).
+// To self-host instead, copy embed.js/widget.js/widget.html from the package
+// tarball into public/webmcp/.
+const RELAY_EMBED_URL =
```
chrisgervang, line 15: Question on the embed since it executes inside the app: have you considered self-hosting it (your comment already documents copying embed.js into `public/webmcp/`), or adding an SRI integrity hash to the script tag? Localhost-only keeps the blast radius to dev machines, though those are the machines holding keys. 

Also do you know if `@mcp-b/global` defers to native `modelContext` when Chrome's origin trial is active, or do the two conflict?

noodles-editor/src/ai-chat/tool-definitions.ts
```diff
@@ -0,0 +1,395 @@
...
+import type { NoodlesProject, SearchCodeParams, ToolResult } from './types'
+
+// JSON Schema subset accepted by both Anthropic tools and WebMCP registerTool
+export interface ToolInputSchema {
+  type: 'object'
+  properties: Record<string, unknown>
+  required?: string[]
+}
+
+export interface ToolDefinition {
```
chrisgervang, line 25: Have you considered a `tier: 'read' | 'write' | 'execute'` field on `ToolDefinition`? If an access-model follow-up happens, the gate becomes data instead of a name list, and it's also the shape the `noodles-mcp` docs server (spec 03 on the #507 branch) would want to import later.

noodles-editor/src/external-control/README.md
```diff
@@ -2,11 +2,13 @@
 
 This module enables external AI tools and scripts to control the Noodles application programmatically through a WebSocket-based API.
 
+> **Note**: For MCP clients (Claude Code, Claude Desktop, Cursor), the recommended path is now **WebMCP** (`src/webmcp/`), which registers the full in-app AI tool surface on `navigator.modelContext` under the same `?externalControl=true` flag — no proxy process required. See "Connecting Claude Code to a Running Browser Instance" in the root `AGENTS.md`. The WebSocket bridge below remains available as a legacy path and for custom non-MCP tooling.
```
chrisgervang, line 5: One thing I'd like your read on, since this note is what makes it official: the WebSocket path keeps its known write-path bug while labeled legacy (worker-bridge never wires the applier, so mutations validate and then drop). Worth fixing, or would you rather remove the path in a follow-up? No strong preference, I just don't want it sitting in limbo where the docs say legacy but the bug says broken.

chrisgervang (approved):
For context: I drafted a WebMCP sub-plan in #507 before this was up, and this PR independently lands a good chunk of its phase 0/1 (shared tool definitions, the project-bridge fix for the write no-op, the dormant context tools). They both pretty much landed on the same shape.

Where the two diverge, this PR mostly look better, and so I'll update the spec to adopt this and focus it on follow ups.

The WebMCP origin trial is live now (Chrome 149 until 156), so I was planning to register noodles.gl for a token and wire it into the build as a follow-up, which makes the hosted app natively agent-visible in ordinary Chrome for the trial window.

noodles-editor/package.json
akre54: Not leftover, annoyingly — it's a build requirement. `@mcp-b/webmcp-ts-sdk` (pulled in by `@mcp-b/global`) declares it as an optional peer but imports it at the top level, so without it installed `vite build` fails with `"zodToJsonSchema" is not exported by "__vite-optional-peer-dep:zod-to-json-schema:..."`. We never hit the code path (it's for their Zod 3 schema conversion; we're on Zod 4 and pass plain JSON Schema), the import just has to resolve. Alternative would be a Vite alias to a stub — happy to switch if you'd rather not carry the dep, but the real package seemed less magical. Added a comment in register.ts pointing at this so it doesn't look orphaned.

noodles-editor/src/webmcp/register.ts
akre54: I went back and forth on this and landed against gating writes: mutation is the point of the integration — an agent that can't edit the graph is a screenshot viewer. And the exposure is already a double opt-in: the user put `?externalControl=true` in the URL *and* installed/configured an agent to reach the tab. A third flag mostly adds friction for the primary (localhost, Claude Code) use case.

The scenario your sketch protects against — a legitimately-installed agent gets prompt-injected and calls `apply_modifications` — is real, but read/write tiers don't actually solve it: a confused agent with read-only tools still exfiltrates the whole project via `get_current_project`, and a write reaches CodeOp (i.e. arbitrary JS) either way. The proportionate version, which I've now pushed to this branch: (a) spec-standard `annotations` (`readOnlyHint`/`destructiveHint`) on all 22 tools, passed through to `registerTool`, so MCP clients do their existing prompt-before-write dance — that's the layer with a human next to it, and Chrome's trial agent will read the same hints; (b) your `window.top !== window` guard in WebMCPProvider.

If hosted exposure during the trial demands more, I suspect the right knob is `safeMode`-shaped (don't execute CodeOps created via external control until unlocked) rather than tool tiers, since the actual risk is code execution, not graph edits. I'll read the 05 sketch with that lens.

noodles-editor/src/webmcp/index.tsx
akre54: Self-hosting is the better fix and I'd take it as a fast follow: three small files into `public/webmcp/`, removes the CDN trust entirely (SRI on the script tag wouldn't cover the `widget.html` the embed fetches at runtime), and works offline. One constraint to write down when we do it: the vendored files must track the relay major — I burned an hour in testing on a 3.x embed + 4.x relay, which fails the discovery handshake silently (zero sources, no errors anywhere).

On native: `@mcp-b/global`'s documented default is `nativeModelContextBehavior: 'preserve'` — if `navigator.modelContext` already exists it wraps the native object and mirrors registrations into it rather than clobbering. That's their claim as of 3.x; worth an explicit check when you wire the trial token, and there's a `window.__webModelContextOptions` escape hatch if it misbehaves.

noodles-editor/src/ai-chat/tool-definitions.ts
akre54: I'd trim `tier` to what actually has a consumer — after chewing on it I don't think we should gate app-side at all (see the register.ts thread), and `'execute'` has no members today; if something eval-shaped ever lands it's a one-line addition then. What I did take from this: the spec's existing vocabulary instead of a new one. Pushed to this branch — `ToolDefinition` now requires an `annotations` field (`readOnlyHint`, plus `destructiveHint`/`idempotentHint` on the four mutating tools), passed through to `registerTool`. Same drift protection you were after (the classification lives on the definition, so a new write tool can't silently pass for read-only), MCP clients already use it to decide prompt-vs-auto-approve, and native Chrome agents will read it. The docs server can import names/schemas without caring about any of this.

noodles-editor/src/external-control/README.md
akre54: Agreed it shouldn't sit in limbo. My lean: make it honest now, remove it later. Cheapest honest state is either (a) wire the WS path's `applyModifications` through the same bridge — genuinely ~10 lines now that `bridge.ts` exists — or (b) have it reject write tools with an explicit "not supported over WebSocket, use WebMCP" error. Since the write path never worked, nobody external can be depending on it, so I'd do (b) and target full removal in the #507 cleanup once WebMCP has some mileage — keeping two write paths bug-compatible isn't worth the maintenance. But if you'd rather it work until removal, (a) is trivial. Your call.

akre54:
Thanks for the thorough pass — and glad the shapes converged with #507; that's reassuring for both. From these threads I've pushed the small hardening to this branch: spec-standard tool annotations (`readOnlyHint`/`destructiveHint`, passed through to `registerTool`) and the top-window guard so embedding pages can't inherit the tool surface. Remaining follow-ups I'd batch into one PR before your origin-trial wiring: self-hosting the relay embed (drops the CDN dependency) and whichever way we go on the legacy WS path (see the README thread).

PR #514, opened 2026-07-08
Derive reference edges in transformGraph for unmounted nodes

chrisgervang:
#### Background

`op()` / mustache references only became executor dependencies when the CodeField editor component mounted and synced `ReferenceEdge`s into React Flow state (`field-components.tsx`). Reference edges are never persisted, and container children aren't mounted while collapsed — so on a fresh project load, a container child whose only upstream link is an `op()` reference executed once before async upstream data (e.g. a FileOp's CSV) arrived and was never re-executed. The result is a silently empty pipeline: `loaded=true`, zero console errors, non-blank screenshot, no data. The bug hides during interactive authoring because opening the container mounts the editor and retroactively wires the reference.

#### Reproduction

**`dev-docs/repros/container-reference-edges/`** — self-contained project (200-row CSV slice). Copy the directory into `noodles-editor/src/examples/repro-ref-edges/` and open `/examples/repro-ref-edges`: the viewer shows `0` forever on `main`, `200` with this fix. The README covers setup, the graph shape, and expected vs actual.

#### Change List
- `transformGraph` now derives reference edges from `op()`/mustache text in string-valued inputs for **every** node on each graph rebuild (`deriveReferenceEdges`), instead of relying on editor components that only mount for visible nodes.
- Derived edges use the exact id/handle shape the CodeField component produces, and both paths dedupe on source+handle+target+handle, so mounted editors keep working unchanged.
- References that don't resolve to an existing node are skipped (no stale-edge errors).
- 3 new unit tests: container-child reference wiring, array-form code + unresolvable-reference skipping, and no-duplicate-against-component-synced-edges. 38/38 pass.
- Reproduction project committed under `dev-docs/repros/container-reference-edges/`.

Verified in-browser: the reference-only container pattern above now produces the full filtered dataset where it previously produced 0 rows.

Related: #515 (GraphInputOp rebuild orphans the container input link) — both were found while verifying eval golden projects that exercise containers headless.

https://claude.ai/code/session_019BbcWNxsAWdwBBPyRNRLQJ


dev-docs/repros/container-reference-edges/data.csv
```diff
@@ -0,0 +1,201 @@
+DateTime,Latitude,Longitude,Depth,Magnitude,MagType,NbStations,Gap,Distance,RMS,Source,EventID
```
akre54, line 1: Is this the earthquake dataset? Could we make the earthquakes example test this feature instead of having a specific example just for testing?

dev-docs/repros/container-reference-edges/noodles.json
```diff
@@ -0,0 +1,92 @@
+{
+  "version": 14,
+  "viewport": {
+    "x": 500,
+    "y": 300,
+    "zoom": 1
+  },
+  "name": "Repro: container child op() reference goes stale",
```
akre54, line 8: My preference is that examples are kept as pure representations of features, rather than tests. Can we move this to the test file instead?

noodles-editor/src/noodles/transform-graph.ts
```diff
@@ -93,12 +94,89 @@ function topologicalSort<N extends Operator<IOperator>>(
...
+  const derived: Array<Record<string, unknown>> = []
+  for (const node of nodes) {
+    for (const [fieldName, value] of Object.entries(node.data?.inputs ?? {})) {
+      const text = Array.isArray(value)
+        ? value.filter((v): v is string => typeof v === 'string').join('\n')
+        : typeof value === 'string'
+          ? value
+          : null
+      // Cheap prefilter — reference syntaxes are `op('...')` and `{{...}}`.
+      if (!text || !(text.includes('op(') || text.includes('{{'))) continue
```
akre54, line 139: I think we need a more robust way of tracking references. Shouldn't this be accessible / already cached?

akre54 (approved):
Approved, with questions

dev-docs/repros/container-reference-edges/data.csv
chrisgervang: Yes, this was a 200-row earthquake-data slice used only to reproduce the executor bug. Follow-up #560 removes this manual repro dataset/project and replaces its behavioral coverage with an inline automated test that verifies both initial execution and a later referenced-source update.

noodles-editor/src/noodles/transform-graph.ts
chrisgervang: There was no model-level reference cache available here. The only cached derivation was the mounted CodeField component useMemo plus its ephemeral React Flow ReferenceEdges; collapsed children never mount and reference edges are not persisted, which was the root cause. deriveReferenceEdges uses the shared getFieldReferences parser at the graph transform boundary and only runs on graph rebuilds, with a cheap syntax prefilter. A future graph-owned reference index could remove the UI/model dual ownership, but it would require coordinated invalidation for edits, loads, renames, relative paths, and field rebuilds; that is a larger refactor and is not required for this correctness fix.

dev-docs/repros/container-reference-edges/noodles.json
chrisgervang: Addressed in follow-up #560. The copy-to-examples repro project is removed, along with the equivalent test-only repro projects from #515 and #516. The #514 scenario now has an inline end-to-end regression in transform-graph.test.ts; the existing focused #515/#516 tests remain in their test files.

chrisgervang: I like the idea of including QA artifacts to PRs for us to validate the feature without needing to set up programs ourselves. This would ideally be added to PR descriptions as some kind of artifact.. I'll see if the github API can attach files instead.

noodles-editor/src/noodles/transform-graph.ts
chrisgervang: The editor’s `useMemo` cache isn't available during fresh loads and inside collapsed containers. We could make the graph model the sole owner of reference dependencies and remove the editor component’s duplicate edge synchronization

akre54: Yeah maybe we move it out of React. I hear you that it's a larger refactor but I'm concerned about computing this state all over the place. If it's needed upstream, let's move it upstream

chrisgervang: #561

PR #519, opened 2026-07-27
Add comprehensive GIS tools and ease-of-use features

akre54:
#### Background

Adds a comprehensive suite of GIS tools to make Noodles competitive with dedicated geospatial workbenches, while preserving the node-graph-first architecture. Every "easy mode" feature produces real nodes that power users can inspect, extend, and animate.

#### Change List
- **Basemap Gallery**: 35 searchable basemap styles with provider/category filtering (Carto, MapTiler, Stadia, ESRI)
- **Data Importer**: URL loading, format auto-detection, coordinate column detection, sample datasets for quick onboarding
- **Layer Panel**: Layer list with visibility/opacity controls, inline Style Panel for per-layer color/radius/width editing
- **Measure Tool**: Geodesic distance and area measurement with unit conversion
- **GeoEditor**: Drawing operator with visual map dialog for creating points, lines, and polygons
- **Format Support**: GeoParquet (via DuckDB Spatial), Shapefile, PMTiles, XYZ Tile operators
- **30+ Geometry/Analysis Operators**: Buffer, Union, Difference, Intersect, Centroid, ConvexHull, Voronoi, Dissolve, Clip, SpatialJoin, PointInPolygon, NearestPoint, Area, Length, clustering (KMeans, DBSCAN), Interpolate, grid generation, transforms (rotate/scale/translate), and more
- **New operator categories**: `geometry` and `analysis` for easy discoverability
- **51 passing tests** covering all new GIS operators
- Homepage and docs repositioned around the animation-studio-for-maps identity


docs/intro.md
```diff
@@ -4,33 +4,42 @@ sidebar_position: 1
 
 # Welcome to Noodles.gl
 
-Noodles.gl is a node-based editor for building geospatial visualizations and animations that are ready to publish — in the browser, no install required.
+Noodles.gl is the animation studio for maps — a node-based editor for building cinematic geospatial visualizations, entirely in the browser.
```
chrisgervang, line 7: Agreed with this positioning - it's the main differentiator compared to all other tools I've used. And, it'd be very very difficult to retrofit on

docs/users/comparison.md
```diff
@@ -1,23 +1,92 @@
-# Tool Comparison
+# How Noodles.gl compares
 
-Comparison with other visualization and animation tools.
+Noodles.gl sits at the intersection of geospatial tools, motion design software, and data visualization libraries. Here's how it relates to tools you may already know.
```
chrisgervang, line 3: I like this tone

docs/users/comparison.md
```diff
@@ -1,23 +1,92 @@
...
+If you need **animated, data-driven map stories** — camera flythroughs, temporal data reveals, parameter-driven transitions exported as video — Noodles.gl is purpose-built for that workflow. Other tools handle pieces of it, but none combine reactive data pipelines, a full keyframe timeline, and GPU-rendered cartography in a single browser-based editor.
 
-- Animated geospatial visualizations
-- Custom data processing pipelines
-- Integration with web applications
-- Timeline-based animations
+## Detailed comparison
 
-## When to Use Alternatives
+### vs. Kepler.gl / Studio
```
chrisgervang: Maybe say FSQ Studio?

docs/users/comparison.md
```diff
@@ -1,23 +1,92 @@
...
-- **Blender**: 3D modeling and complex scene creation
+Kepler is excellent for fast geospatial exploration — drop a CSV, see points on a map, adjust filters. But it produces **static views**. There's no timeline, no keyframeable parameters, no way to animate a camera move or export a video. When you need the visualization to *move*, Kepler is the starting point you outgrow.
+
+| | Kepler.gl | Noodles.gl |
+|---|---|---|
+| Drop CSV and explore | Instant | Instant (via Import Data) |
+| Animate camera | No | Keyframeable with curves |
+| Animate data over time | Time filter widget only | Full timeline, any parameter |
+| Composable pipeline | No | Reactive node graph |
+| Export video | No | MP4, any resolution |
```
chrisgervang: "No" might need an asterisk. It's there but limited to exporting a video of the time filter. Only present resolutions.

docs/users/comparison.md
```diff
@@ -1,23 +1,92 @@
...
+| Custom code | Limited expressions | Full JavaScript, DuckDB SQL |
```
chrisgervang: They have DuckDB sql in Kepler. The expressions are only on enterprise tier Studio, which is out of reach for most readers

docs/users/comparison.md
```diff
@@ -1,23 +1,92 @@
...
+After Effects is the gold standard for motion graphics — but it has no concept of geospatial data, map projections, or reactive data flow. To animate a map in AE, you'd export frames from a GIS tool, import them as assets, and manually keyframe everything. If the data changes, you start over.
+
+Noodles.gl gives you AE-style keyframing (bezier curves, timeline scrubbing, easing) applied directly to a live data pipeline. Change the underlying query and every frame of your animation updates automatically.
+
+| | After Effects | Noodles.gl |
+|---|---|---|
+| Keyframe timeline | Full, professional | Full, with bezier curves |
+| Data-driven | Manual import | Live reactive pipeline |
+| Map projections | None (raster only) | Globe, Mercator, Orthographic |
+| 40+ geo layer types | No | Arc, hex, heatmap, trips, etc. |
```
chrisgervang: The fair comparison is AE + GeoLayers, which costs more and isn't as capable. But in the niche, it's widely used

docs/users/comparison.md
```diff
@@ -1,23 +1,92 @@
...
+| Custom visualization | Unlimited (code) | 40+ layers + CodeOp for custom |
+| Visual editor | No | Full node graph + timeline |
+| GPU rendering | No (SVG/Canvas) | WebGL via Deck.gl |
+| Large datasets (1M+ rows) | Difficult | Handled natively |
+| Animation | Code your own | Built-in timeline |
+| Shareable projects | Deploy as app | JSON file or embed |
+
+### vs. QGIS / GeoLibre (desktop GIS)
+
+Desktop GIS tools are unbeatable for spatial analysis — hundreds of algorithms, format support, and editing capabilities. But they produce static maps. Noodles.gl is not trying to be a GIS workbench. It's the tool you reach for *after* analysis, when you need to present results with motion and narrative.
```
chrisgervang: Good framing. And we might consider how to make import after analysis as easy as possible

docs/users/comparison.md
```diff
@@ -1,23 +1,92 @@
...
+
+| | QGIS / GeoLibre | Noodles.gl |
+|---|---|---|
+| Spatial analysis algorithms | Hundreds | CodeOp + DuckDB |
+| Format support | 30+ | GeoJSON, CSV, Parquet, SQL |
+| Geometry editing | Full | Coming soon (GeoEditor) |
+| Animation / timeline | None | Core feature |
```
chrisgervang: GeoLibre has deck trips layer, so there's some limited support

docs/users/comparison.md
```diff
@@ -1,23 +1,92 @@
...
+| Data-driven styling | Static rules | Reactive, animated |
+| Browser-based | GeoLibre yes, QGIS no | Yes |
+
+### vs. Flourish / Datawrapper
+
+Template-based visualization tools are easy to use but constrained. You get pre-built chart types with limited customization. Noodles.gl has no templates — instead, you compose visual pipelines from primitive operators, which means the ceiling is much higher.
+
+| | Flourish | Noodles.gl |
+|---|---|---|
+| Time to first visualization | Minutes | Minutes |
```
chrisgervang: This feels aspirational.. maybe Minutes in noodles with a well tuned AI agent or when using a template, otherwise it's quite a bit to learn

docs/users/comparison.md
```diff
@@ -1,23 +1,92 @@
...
+| Customization ceiling | Low (templates) | Unlimited (composable graph) |
+| Map animations | Preset transitions | Any parameter, any curve |
+| Custom data transforms | No | CodeOp, DuckDB, expressions |
+| Self-hosted | No | Yes (static files) |
+| Export video | Limited | Full MP4 |
+
+## When to use Noodles.gl
```
chrisgervang, line 79: These are good

chrisgervang (commented):
Leaving doc comments to start. It'll be easier to review this as a split set of PRs (many seem independent anyways)

akre54:
Thanks Chris, all six of your notes are in:

- Heading is now "vs. Kepler.gl / Foursquare Studio".
- Video export row says "Time filter only, preset resolutions" instead of "No".
- Custom code row says "DuckDB SQL; expressions on Studio Enterprise".
- Retitled the AE section to "After Effects + GeoLayers" and noted it's widely used and capable, just a paid plugin on a subscription with static data imports.
- GeoLibre animation row is now "Trips layer playback (GeoLibre)" instead of "None".
- Time to first visualization now reads "Minutes from a template or with the AI assistant; longer if you start from a blank graph".

Also updated two rows that went stale in this PR: geometry editing is "Basic (GeoEditor)" rather than "coming soon", and format support lists Shapefile, GeoParquet and PMTiles.

On making import-after-analysis easy: the importer detects GeoJSON, CSV, Shapefile, GeoParquet and PMTiles and scaffolds the whole pipeline, but you still have to open it first. Dropping a file straight onto the canvas is the obvious next step. I'll track that separately rather than grow this PR.

Greptile's three P1s are fixed in the same commit:

1. GeoParquetOp interpolated the URL and geometry column into raw SQL, so any value with a quote in it broke the query. Added `sqlLiteral` and `sqlIdentifier` helpers in `src/utils/sql.ts` with tests, and floored the LIMIT.
2. The GeoEditor accepted a polygon after two clicks and closed it into a 3-position ring. Polygons now need 3 points, and the Finish button and point counter use the same rule so the UI can't offer an invalid shape.
3. Dialog features were initialized once at mount, so reopening could save stale geometry over a newer upstream value. It now re-reads the operator each time it opens.

akre54:
Pushed. Drop-to-import is in this PR after all, rather than deferred, plus two more commits.

**Drop-to-import** (`6e79b408`): drag a `.csv`, `.geojson`, `.json`, `.shp`, `.parquet` or `.pmtiles` onto the graph and the pipeline gets built at the drop point, attached to the existing renderer or to a new basemap and renderer if there isn't one. The source node ends up selected and framed. Multiple files each get their own row, and unsupported files show a dismissible error instead of the browser navigating away from the app.

Getting there meant pulling the dialog's pipeline construction out into `import-pipelines.ts`, which has no React or store dependency, and moving the write-then-scaffold sequence into a `useFileImport` hook shared by the dialog and the canvas. That surfaced three bugs worth calling out:

1. The CSV pipeline wired three edges to a `/deck` node it never created, so importing a CSV into an empty project gave you a ScatterplotLayer connected to nothing. The GeoJSON path created the renderer, the CSV path didn't. There's now a test asserting every edge endpoint is a node the builder actually returns.
2. BoundingBox was wired to GeoJSON sources, but its input is `ArrayField(Point2DField)` and it can't read a FeatureCollection. It's wired for tabular sources only now.
3. Shapefile, GeoParquet and PMTiles existed as operators but weren't reachable as imports. All three have detection and a pipeline now, and binary formats round-trip as a Blob rather than going through `file.text()`, which would have corrupted them.

I also deleted `data-importer-tool.test.ts`. It re-implemented the builder inline and asserted against its own copy, so it would have kept passing while the `/deck` bug shipped. Replaced with 19 tests against the real module.

**Shelf entry points** (`43b6d0cc`): the ~35 GIS operators had no way in other than the generic node search. Added a declarative recipe registry (34 recipes across geometry, combine, analysis, grid, transform, source), one generic wizard that drives any of them, and a searchable grouped dropdown. The wizard populates its input picker from every GeoJSON or data output already in the graph, so you don't have to know handle names. Everything it asks becomes a serialized operator input and the parameter node is left selected, so it's clear where to change values or add keyframes afterwards. One test instantiates every recipe's operator and asserts each declared input, param and output key exists, which is what caught the handle-name drift (`feature` vs `geojson`, `out.inside`, `out.sourceConfig`).

Full suite is 2509 passing, build is clean, and biome reports nothing on any file this PR touches.

noodles-editor/src/noodles/operators.ts
```diff
@@ -6902,6 +6902,1246 @@ export class GeoJsonOp extends Operator<GeoJsonOp> {
...
+    limit,
+  }: ExtractProps<typeof this.inputs>): Promise<ExtractProps<typeof this.outputs>> {
+    if (!url) return { featureCollection: { type: 'FeatureCollection', features: [] }, data: [] }
+    const db = await duckDbInstance
+    const conn = await db.connect()
+    try {
+      await conn.query('INSTALL spatial; LOAD spatial;')
+      const limitClause = limit > 0 ? `LIMIT ${limit}` : ''
+      const result = await conn.query(
+        `SELECT ST_AsGeoJSON(${geometryColumn}) as geojson, * EXCLUDE(${geometryColumn}) FROM read_parquet('${url}') ${limitClause}`
```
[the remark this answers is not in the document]
akre54: Valid. Fixed in 9a52a7f1: added `src/utils/sql.ts` with `sqlLiteral`/`sqlIdentifier` (SQL standard quote doubling) since DuckDB-WASM has no parameter binding for table functions like `read_parquet()`. The GeoParquet query now quotes both the URL and the geometry column, and `limit` goes through `Math.floor`. Covered by 7 tests in `src/utils/sql.test.ts`, including an injection attempt.

noodles-editor/src/noodles/components/geo-editor-dialog.tsx
```diff
@@ -0,0 +1,240 @@
...
+        setFeatures(prev => [...prev, newFeature])
+      } else {
+        setCurrentPoints(prev => [...prev, point])
+      }
+    },
+    [drawMode]
+  )
+
+  const finishShape = useCallback(() => {
+    if (currentPoints.length < 2) return
```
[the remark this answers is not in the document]
akre54: Valid. Fixed in 9a52a7f1: added `minPointsFor(mode)` (3 for polygon, 2 otherwise) and applied it in all three places, so the Finish button does not render until the shape is valid and the guard rejects it if it somehow is.

noodles-editor/src/noodles/components/geo-editor-dialog.tsx
```diff
@@ -0,0 +1,240 @@
...
+    return []
+  }
+}
+
+export function GeoEditorDialog({ open, onOpenChange, operator }: GeoEditorDialogProps) {
+  const [drawMode, setDrawMode] = useState<DrawMode>('polygon')
+  const [currentPoints, setCurrentPoints] = useState<[number, number][]>([])
+  const [features, setFeatures] = useState<Feature[]>(() =>
+    parseGeoJson(String(operator.inputs.geojson?.value ?? ''))
+  )
```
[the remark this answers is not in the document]
akre54: Valid. Fixed in 9a52a7f1: the one-time `useState` initializer is replaced with a `useEffect` keyed on `[open, operator]`, so the dialog re-reads the operator on every open instead of keeping whatever was on screen last time.

PR #529, opened 2026-08-02
Rename generic node IDs to semantic names in examples

akre54:
#### Background
Example projects contained generic auto-generated node IDs (e.g., `duckdb-0`, `accessor-2`) that made the node graphs harder to read and understand. This PR renames all generic node IDs to semantic names that describe what each node does.

#### Change List
- Renamed 14 generic node IDs across 5 example projects (chargers, orbit, nyc-census, world-flights, custom-maplibre-layer-test)
- Updated all edge references and timeline keys to use new semantic names
- Regenerated website screenshots for chargers and world-flights examples
- Added `validate-examples.js` script with pre-commit hook integration to enforce semantic naming
- Added `setup-hooks.sh` script for easy git hook installation


package.json
chrisgervang, line 1: This unintentionally reformatted the file to use tabs instead of spaces. npm intentionally formats a lockfile to match package.json whenever it updates the lockfile. Pushed up #531 to fix

PR #530, opened 2026-08-05
test: add Playwright visual regression tests for examples

akre54:
#### Background

Visual regression testing helps catch unintended rendering changes across all example projects. Previously, changes to operators, layers, or rendering could break examples without detection.

#### Change List
- Add Playwright E2E visual regression tests with dynamic example discovery
- Detect animated examples by parsing keyframes, test multiple frames (0s, 0.5s, 1s, 2s)
- Commit 11 baseline snapshots for CI comparison (platform-specific)
- Expose `window.deck` and `window.getTimelineStore` for test access
- Add comprehensive test README with setup, troubleshooting, CI integration
- Configure Playwright with 150s timeout for slow-loading examples
- Gitignore test artifacts (playwright-report/, test-results/) but keep snapshots


chrisgervang (commented):
Why are we testing the UI+output instead of just the output? UI could change in a valid way that breaks the output, and it'll be difficult to tell.

PR #532, opened 2026-08-06
fix: silence already-applied Vitest postinstall patch

chrisgervang:
## Summary

Stop the `@vitest/browser-playwright` postinstall script from warning when its workaround is already present.

After the first successful patch, later installs could no longer find the original source and printed `target not found — patch may be outdated`. That warning was misleading: the target was missing because the patch had already been applied.

This change checks for the exact injected handler and exits silently when it is found. The existing warning remains for genuinely unknown upstream source. The underlying `route.fulfill()` workaround is unchanged.

## Testing

- `node --check scripts/patch-vitest-browser-playwright.cjs`
- Ran the script against an already-patched installation and confirmed it produced no output.

## Context

The workaround was introduced in #405 for a Vitest/Playwright teardown race.


chrisgervang:
Closed in favor of #538

PR #533, opened 2026-08-06
chore: upgrade deck.gl to 9.3.8

### LINKED ISSUE #528: MapLibre `antialias` and `preserveDrawingBuffer` are silently ignored on maplibre-gl v5
## Summary

`noodles-editor/src/timeline-editor.tsx` passes `antialias` and `preserveDrawingBuffer` as **top-level** MapLibre map options. On maplibre-gl v5 (we pin `^5.21.1`) both moved into `canvasContextAttributes`, and the top-level keys are no longer read at all. Both settings are currently no-ops.

The visible consequence is that MSAA is off, so deck.gl layers interleaved into the basemap render with no antialiasing whatsoever — deck.gl's `PathLayer` has no analytic antialiasing in its fragment shader and depends entirely on the default framebuffer's MSAA. Basemap lines stay crisp (MapLibre computes analytic coverage in `line.fragment.glsl`), while deck.gl strokes look aliased next to them. Most visible on high-resolution exports.

## Evidence

`maplibre-gl@5` builds its context attributes only from `canvasContextAttributes`:

```ts
// maplibre-gl/src/ui/map.ts
defaultOptions = { canvasContextAttributes: {          // :500
    antialias: false, preserveDrawingBuffer: false, ... } }

const resolvedOptions = {...defaultOptions, ...options, canvasContextAttributes: {   // :736
    ...defaultOptions.canvasContextAttributes, ...options.canvasContextAttributes }}

this._canvasContextAttributes = {...resolvedOptions.canvasContextAttributes};        // :778
```

`options.antialias` and `options.preserveDrawingBuffer` are never read. The top-level keys survive into `resolvedOptions` and are then discarded.

### 1. Move both into `canvasContextAttributes`

`noodles-editor/src/timeline-editor.tsx`, around line 179:

```diff
 const mapProps: MapProps = {
   interactive: false,
-  antialias: true,
-  preserveDrawingBuffer: true,
+  canvasContextAttributes: {antialias: true, preserveDrawingBuffer: true},
   onLoad: ({ target: map }) => {
```

`preserveDrawingBuffer` matters independently of the aliasing — the screenshot and video capture paths in `src/render/` rely on it.

### 2. Drop `useDevicePixels` from `deckProps`

`noodles-editor/src/timeline-editor.tsx`, around line 147. It has no meaning in interleaved mode —
the base map owns the canvas and its drawing buffer size.

Worse, on `deck.gl@9.2` it actively leaks. `MapboxOverlay.filterProps` reads `interleaved` off the
*incoming* props, so while the constructor call is safe (`new MapboxOverlay({...props, interleaved:
true})`), the later `deck.setProps({...props, onAfterRender})` in `DeckGLOverlay` omits
`interleaved`. It defaults to `false`, and `useDevicePixels: false` reaches `Deck`, which forwards it to `_canvasContext.setProps(...)` on MapLibre's shared canvas. On a HiDPI display that clamps the
shared drawing buffer to 1× CSS and halves effective resolution for the basemap and deck alike.
No-op on 1× displays.

Fixed upstream in [visgl/deck.gl#10510](https://github.com/visgl/deck.gl/pull/10510), which is on `master` but not in any published release as of `9.4.0-alpha.2`. Removing the prop fixes it on 9.2 today regardless.

## Notes

Even with MSAA correctly enabled, multisampling is expensive at these export resolutions (a
3840×2160 CSS canvas at `devicePixelRatio: 2` means a 7680×4320 multisampled buffer) and still quantizes coverage for sub-pixel strokes. deck.gl 9.4 might add a `antialiasing` prop to `PathLayer` and `LineLayer` that computes coverage analytically in the shader, which would let this app turn MSAA back off.

chrisgervang:
Closes #528

#### Background

Upgrade deck.gl from 9.2.11 to 9.3.8. This release fixes device-pixel handling for interleaved MapLibre rendering and removes the legacy FPS widget in favor of the consolidated Stats widget.

MapLibre GL v5 also reads antialiasing and frame-capture settings from canvasContextAttributes, so the existing top-level options were ignored.

#### Change List

- Upgrade deck.gl to 9.3.8 and align the directly imported @luma.gl/effects package at 9.3.6.
- Configure antialiasing and preserveDrawingBuffer through MapLibre canvas context attributes.
- Keep useDevicePixels disabled for standalone Deck rendering; deck.gl 9.3.8 now ignores it automatically for an interleaved MapboxOverlay.
- Preserve FpsWidgetOp project compatibility while rendering deck.gl's replacement StatsWidget.
- Update built-in widget descriptors to use the constructors exported by deck.gl 9.3.
- Add regression coverage for rendering defaults and runtime widget-constructor resolution.

#### Testing

- npm test --workspace noodles-editor -- --run (2,679 passed, 10 skipped)
- npm run lint --workspace noodles-editor
- npm run format:check --workspace noodles-editor
- npm run build --workspace noodles-editor


chrisgervang:
Recent renders have come out blurry, quick fix to improve quality

akre54 (approved):
Fine by me. The logic is a little spaghetti now but I like the idea of extracting the defaults to a separate file.

Minor note inline but fine to merge as-is

noodles-editor/src/timeline-editor.tsx
```diff
@@ -140,15 +141,7 @@ export default function TimelineEditor() {
...
-    deviceProps: {
-      type: 'webgl',
-      powerPreference: 'high-performance',
-      webgl: {
-        stencil: true,
-      },
-    },
-    useDevicePixels: false,
-    ...visualization.deckProps,
+    ...createDeckRenderingProps(visualization.deckProps, basemapEnabled),
```
akre54: Maybe slight pref towards doing this inline here, instead of a helper function:

```ts
const deckProps: DeckProps = {
  ...deckRenderingDefaults
  ...visualization.deckProps,
  ...(basemapEnabled ? {} : { useDevicePixels: false })
}
```

IMO reads a little cleaner without the indirection

chrisgervang: We could inline it but this snippet isn't equivalent. If we went back to this code,

- With a basemap, `visualization.deckProps.useDevicePixels` would survive the spread and could reintroduce the interleaved-mode bug.
- Without a basemap, the final {useDevicePixels: false} would override an explicit standalone setting such as true.

The helper deliberately strips the property in interleaved mode, so our inline version should too. The inline version looks like:

```ts
const deckProps: DeckProps = {
  ...deckRenderingDefaults,
  useDevicePixels: false,
  ...visualization.deckProps,
  onDeviceInitialized: device => {
    visualization.deckProps?.onDeviceInitialized?.(device)
    redraw()
  },
  onAfterRender: () => {
    // Existing callback body
  },
}

// MapLibre owns the shared canvas drawing buffer in interleaved mode.
if (basemapEnabled) {
  delete deckProps.useDevicePixels
}
```

chrisgervang: It'd be nice to get TimelineEditor into a easier to test state, or just invest in making complex mocks for it. I'm a bit concerned of inlining this right now because we can't easily regression test it in the TimelineEditor machinery

akre54: What's the actual code that checks `useDevicePixels`? If it doesn't use an `in` check, i.e. `if ('useDevicePixels' in props) { ...` then we could just set `{ useDevicePixels: undefined }` if `basemapEnabled`

But also fine. I think it's a little more obvious to keep the props organized but this is very minor

chrisgervang: Even better: Spoke to Xiaoji and we decided to ship a patch in deck 9.3 instead of 9.4 so now it should be safe to always define useDevicePixels. Internally, it'll drop it for interleaved renderers

https://github.com/visgl/deck.gl/pull/10510

akre54: Nice!

chrisgervang:
@greptile-apps re-review

PR #538, opened 2026-08-06
test: remove redundant global fields mock

chrisgervang:
## Summary

- remove the no-op global `vi.mock("./noodles/fields")` factory from the Vitest setup file
- retain the existing `@vitest/browser-playwright` postinstall workaround
- keep dependency versions and Vitest configuration unchanged

## Refined root cause

Vitest setup files already run before test-module imports, so the fields passthrough mock was unnecessary. In Browser Mode, that global factory mock registered a manual Playwright route for every test file, adding 131 routes and making the provider teardown race much more likely.

It was an amplifier, not the sole trigger. The suite still contains manual factory mocks in individual tests. After merging latest main, one clean unpatched local run reproduced `route.fulfill: Route is already handled!` even without the fields mock. The provider workaround therefore remains necessary.

## Experiment evidence

- unpatched control with the global mock: 8/10 passed; runs 4 and 7 failed with the target route error and unhandled rejection
- initial unpatched treatment without the global mock: 10/10 passed; average setup fell from 21.99s to 4.19s
- historical CI treatment on `dd63636a`: 10/10 `test-app` attempts passed with no target signatures
- post-merge unpatched treatment: reproduced the same race once, demonstrating residual exposure from other manual mocks

## Validation of current safe change

- [x] merged `origin/main` at `8033989929f9fde2556dac79093e9f303e508924`
- [x] retained the provider patch and confirmed postinstall applies it
- [x] local suite: 130/130 files, 2,663 tests passed
- [x] `npm run lint --workspace noodles-editor`
- [x] `npm run format:check --workspace noodles-editor`
- [x] CI for current head `1dcaf166`: all jobs passed ([run 31153089914](https://github.com/joby-aviation/noodles.gl/actions/runs/31153089914))

## Earlier red workflow

The first workflow attempt failed only in `test-website`: GitHub cache restore returned HTTP 400, then `npm ci` failed with `EEXIST` in `/home/runner/.npm/_cacache/tmp`. In that same attempt, `test-app` and `lint-format` passed. Later attempts were fully green: https://github.com/joby-aviation/noodles.gl/actions/runs/31132430259/attempts/1


chrisgervang:
Had AI run CI 10 times on this to see if it'd flap, and all passed. I think we can get rid of this patch

chrisgervang:
Dang, it hit the race condition locally. We can still remove the unnecessary mock, but I suppose the patch should stay.

akre54 (approved):

PR #539, opened 2026-08-07
Fix duplicate for loop group IDs

chrisgervang:
#### Background
ForLoop visual groups used operator-store ID generation even though group nodes are not operators, causing multiple loops to reuse the same serialized group ID.

#### Change List
- Derive each group ID from its unique ForLoopBegin operator ID so their numeric suffixes stay aligned.
- Add regression coverage verifying multiple loops use distinct groups and correct parent references.

Tests: `npm test -- --run src/noodles/__tests__/node-operations-integration.test.tsx src/noodles/hooks/use-node-drop-on-edge.test.ts src/noodles/utils/serialization.test.ts`; `npm run lint`; `npm run format:check`.


akre54 (approved):
👍

PR #540, opened 2026-08-07
Add reusable group layout reconciliation

chrisgervang:
## Summary

Adds the coordinate and bounds foundation used by the ForLoop visual-membership follow-up. This PR is stacked above the merged #539 and intentionally contains no editor wiring.

## Changes

- Reparent React Flow nodes without changing their absolute canvas position.
- Fit groups around measured direct children with padding and minimum dimensions.
- Lay out nested groups from the inside out and keep parents before children in node order.
- Defer shrinking while any child is still unmeasured.
- Keep fitted `style`, `width`, `height`, and `measured` dimensions synchronized so stale React Flow caches cannot override new bounds.
- Serialize group dimensions canonically in `style` instead of persisting transient measurement caches.

### Automated

- Full editor suite: 2,683 passed, 10 skipped.
- Regression coverage for stale cached dimensions both before and after children are measured.
- `npm run lint`
- `npm run format:check`
- `npm run build`

### Fixture validation

- The supplied migrated project normalizes its oversized group from `12287×7652` to `3083×1393` and its over-tall group from `1200×682` to `1223×437`.

## Documentation

No documentation changes needed.

## Related PRs

- Followed by #541, #542, and migration PR #544.

PR #542, opened 2026-08-07
Align ForLoop execution scope with visual membership

chrisgervang:
## Summary

Uses the same paired Begin-to-End path definition for loop execution that the preceding visual-membership PR uses for rendering. This is the execution layer in the stack above #539.

## Changes

- Pass serialized visual nodes and edges to `updateGraph`, which derives and installs visual-group Begin/End/Meta definitions centrally.
- Restrict execution scope to nodes that are downstream of Begin and upstream of the paired End.
- Exclude downstream dead-end branches that cannot contribute to End.
- Pair nested loop boundaries from their visual groups instead of stopping at the first reachable End.
- Execute nested loops atomically within each parent iteration so their boundaries and internal operators are not pulled twice.
- Remove unused execution-scope `groupId` state and path-based group inference.

### Automated

- `npm test -- --run` — 2,682 passed, 10 skipped.
- Regression coverage for dead-end branches, nested boundary pairing, and nested-loop execution counts.
- `npm run lint`
- `npm run format:check`
- `npm run build`

### Manual

1. Insert an operator into a ForLoop Begin-to-End chain and provide an input array.
   - Expected: the inserted operator runs once per item and the End output collects its results.
2. Add a branch from Begin that does not reconnect to End.
   - Expected: the dead-end branch is not part of loop iteration scope.
3. Place a complete inner ForLoop on the outer Begin-to-End path.
   - Expected: each Begin is paired with its visual group End, and the inner loop runs once per outer item.

## Documentation

No documentation changes needed.

## Related PRs

- Depends on #541.
- Earlier stack layers: #540 and #539.
- Followed by migration PR #544.

### DIFF
--- a/noodles-editor/src/noodles/graph-executor.ts
+++ b/noodles-editor/src/noodles/graph-executor.ts
@@ -6,7 +6,11 @@ import { visibilityAdaptiveLoop } from '../utils/worker-timer'
 import type { Field } from './fields'
 import type { ForLoopBeginOp, ForLoopEndOp, ForLoopMetaOp, IOperator, Operator } from './operators'
 import { getAllOps } from './store'
-import type { ForLoopDefinition } from './utils/for-loop-group-utils'
+import {
+  type ForLoopDefinition,
+  findForLoopDefinitions,
+  type GraphNode,
+} from './utils/for-loop-group-utils'
 import type { OpId } from './utils/id-utils'
 
 export type ComputeResult<T = unknown> = {
@@ -140,7 +144,6 @@ type ForLoopScope = {
   endOp: ForLoopEndOp
   metaOp?: ForLoopMetaOp
   scopeNodeIds: string[]
-  groupId: string
 }
 
 // GraphExecutor - manages execution of the operator graph
@@ -650,7 +653,6 @@ export class GraphExecutor {
       endOp,
       metaOp,
       scopeNodeIds,
-      groupId: allScopes?.find(scope => scope.beginOp === beginOp)?.groupId ?? '',
     }
     const loopScopes = allScopes ?? this.findForLoopScopes()
     const nestedScopes = this.findDirectNestedForLoopScopes(currentScope, loopScopes)
@@ -852,7 +854,6 @@ export class GraphExecutor {
             endOp,
             metaOp,
             scopeNodeIds,
-            groupId: definition?.groupId ?? (op.id.split('/').slice(0, -1).join('/') || '/'),
           })
         }
       }
@@ -1020,7 +1021,7 @@ export function stopExecutor(): void {
 }
 
 // Update graph from edges - syncs nodes from the store
-export function updateGraph(edges: Edge[], forLoopDefinitions: ForLoopDefinition[] = []): void {
+export function updateGraph(nodes: GraphNode[], edges: Edge[]): void {
   if (!globalExecutor) {
     initializeExecutor()
   }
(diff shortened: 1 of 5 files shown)


noodles-editor/src/noodles/transform-graph.ts
```diff
@@ -204,7 +205,7 @@ export function transformGraph<
   })
 
   // Update dependency graph
-  updateGraph(edges as unknown as ExecutorEdge[])
+  updateGraph(edges as unknown as ExecutorEdge[], findForLoopDefinitions(nodes))
```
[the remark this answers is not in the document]
chrisgervang: Fixed in 3643ba00. transformGraph now passes the original serialized node list to definition discovery, including visual group nodes. Added an integration regression that transforms nested loop groups and verifies each Begin is paired with its own End in the executor.

chrisgervang:
@greptile-apps re-review

noodles-editor/src/noodles/graph-executor.ts
```diff
@@ -730,50 +737,61 @@ export class GraphExecutor {
...
+              (candidate?.constructor as { displayName?: string } | undefined)?.displayName ===
+              'ForLoopEnd'
+          )
+        const endOp = (configuredEnd ?? inferredEnd) as ForLoopEndOp | undefined
 
         if (endOp) {
+          const upstream = reachable(endOp.id, this.upstream)
+          const scopeNodeIds = [...this.nodes.keys()].filter(
+            id => downstream.has(id) && upstream.has(id)
+          )
```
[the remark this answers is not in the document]
chrisgervang: Fixed in 169083e4. executeFrame now schedules only root loop scopes, and each nested loop executes recursively as one atomic step inside every parent iteration. Parent scopes skip the nested boundaries and internal operators instead of pulling them again. Added an execution regression that maps two nested arrays and verifies the inner scope runs exactly once per outer item.

noodles-editor/src/noodles/transform-graph.ts
```diff
@@ -204,7 +205,7 @@ export function transformGraph<
   })
 
   // Update dependency graph
-  updateGraph(edges as unknown as ExecutorEdge[])
+  updateGraph(edges as unknown as ExecutorEdge[], findForLoopDefinitions(_nodes))
```
akre54: Does this need to be top-level?

akre54: Should we instead pass nodes and edges, and have `updateGraph` process the forLoopDefinitions?

noodles-editor/src/noodles/graph-executor.ts
```diff
@@ -714,66 +771,88 @@ export class GraphExecutor {
...
+            )
+          const metaOp = (configuredMeta ?? inferredMeta) as ForLoopMetaOp | undefined
+
           scopes.push({
             beginOp: op as ForLoopBeginOp,
             endOp,
             metaOp,
             scopeNodeIds,
-            groupId: op.id.split('/').slice(0, -1).join('/') || '/',
+            groupId: definition?.groupId ?? (op.id.split('/').slice(0, -1).join('/') || '/'),
```
akre54: Has this logic for groupId shifted? Should we just make the definition the source of truth instead of parsing it in two places?

Also this split logic seems messy. We should've extracted it out to a helper

akre54 (approved):
Looks like a good fix. Couple stylistic q's

noodles-editor/src/noodles/transform-graph.ts
chrisgervang: Addressed in b74fc45e. The full top-level node list is required because visual group nodes are filtered out before operator instantiation, but transformGraph now passes nodes and edges directly and updateGraph owns deriving and installing the loop definitions.

noodles-editor/src/noodles/graph-executor.ts
chrisgervang: Addressed in b74fc45e. groupId was not used by execution, so I removed it from ForLoopScope and deleted the path-splitting fallback entirely. The visual definition remains the source of truth for Begin/End/Meta pairing.

PR #544, opened 2026-08-07
fix: migrate legacy for loop groups

chrisgervang:
#### Background

Projects created before ForLoop body IDs became unique can contain multiple valid loop chains but only one oversized visual group, with the loop markers and body nodes left unparented. Without visual definitions, nested scope pairing and layout cannot be repaired reliably at runtime.

#### Change List

- Add schema migration 016 to identify legacy Begin and End operators by type and pair them by directed graph reachability, independent of node names.
- Pair nested loops from the inside out with the nearest unambiguous reachable End; leave incomplete, disconnected, or ambiguous boundaries unchanged.
- Reuse or create a unique `/for-loop-body-N` group for each confirmed pair, associate its Meta operator, and derive body membership from the connected subgraph.
- Preserve absolute node positions and advance bundled examples to schema version 16.

#### Test Plan

- Migration regression modeled on a legacy project with three renamed loop triplets, one oversized group, and no parent memberships.
- Verify nested loops pair inside-out and incomplete or disconnected boundaries are not guessed.
- Validate the attached 73-node project reconstructs three groups with the expected connected children and unchanged canvas positions.
- Full editor test suite, lint, format check, and production build.

### DIFF
--- a/noodles-editor/src/noodles/utils/for-loop-group-utils.ts
+++ b/noodles-editor/src/noodles/utils/for-loop-group-utils.ts
@@ -11,13 +11,6 @@ export type ForLoopDefinition = {
   metaIds: string[]
 }
 
-const LOOP_BEGIN_SUFFIX = /for-loop-begin(-\d+)?$/
-
-function relatedLoopId(beginId: string, part: 'body' | 'end' | 'meta'): string | undefined {
-  if (!LOOP_BEGIN_SUFFIX.test(beginId)) return undefined
-  return beginId.replace(LOOP_BEGIN_SUFFIX, `for-loop-${part}$1`)
-}
-
 type LoopGroup = {
   groupId: string
   beginId: string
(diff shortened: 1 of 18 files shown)


noodles-editor/src/noodles/utils/for-loop-group-utils.ts
```diff
@@ -11,6 +11,13 @@ export type ForLoopDefinition = {
   metaIds: string[]
 }
 
+const LOOP_BEGIN_SUFFIX = /for-loop-begin(-\d+)?$/
```
chrisgervang: This should be by op type and walking connections rather than name to be robust

chrisgervang: Fixed in 18fcde42 (restacked after the #542 review updates). The migration now discovers boundaries by ForLoop operator type and walks directed connections, pairing innermost loops first with the nearest unambiguous reachable End. Marker names are no longer used; renamed and nested-loop regressions cover the behavior.

PR #555, opened 2026-08-18
feat: add switchable vector input ports

chrisgervang:
## Why

Geocoder and GeoJSON Point outputs already represent a geographic point, but MapViewState previously required separate longitude and latitude inputs. Users either had to add adapter nodes or discard the semantic Point type.

This change lets a vector input match the way its source data is shaped without creating another operator.

## User experience

- MapViewState now has a Point2D `center`, with zoom, pitch, and bearing remaining independent.
- Whole-value mode connects Geocoder or GeoJSON Point outputs directly to `center`.
- Component mode exposes `lng` / `lat`; connected components are driven while unconnected siblings remain editable and animatable.
- Point Lookup is available in both editable layouts.
- The layouts are mutually exclusive. Switching is disabled until parent and component connections are disconnected, avoiding priority ambiguity.
- Keyframe controls remain in the Properties Panel sidebar and never appear on graph nodes.

Whole-value and component layouts

## Framework mechanism

Point2D, Point3D, Vec2, and Vec3 inputs can now expose either a parent handle (`par.center`) or numeric channel handles (`par.center.lng`, `par.center.lat`). The selected mode is stored per node in `data.inputPortModes`.

Synchronized channel fields preserve object and tuple return types and compose mixed constant/accessor channels for accessor-capable vectors. Graph resolution, validation, execution, cleanup, serialization, history, connection dimming, and programmatic graph edits understand channel handles. Outputs, Vec4, colors, and arbitrary compound fields are unchanged.

## Migration and compatibility

Migration 018 converts MapViewState longitude/latitude values, connections, visibility, and timeline paths to `center.lng` / `center.lat`, with a down migration included. Existing nodes retain their component layout, animation data, static overrides, and markers. Deprecated runtime aliases keep `par.longitude` and `par.latitude` CodeOp references reactive. Bundled projects are migrated to version 18.

## Alternatives considered

- **PointViewState:** duplicates MapViewState and expands the operator vocabulary.
- **SplitLngLat:** adds wiring and loses the semantic Point connection.
- **Show parent and component inputs together:** requires precedence rules for competing writers.
- **MapViewState-only behavior:** leaves the same UX problem on other vector inputs.

## Documentation

- [Vector Input Ports — user guide](https://github.com/joby-aviation/noodles.gl/blob/58e6dfd78c5c68461dc27b8b08adec804ff85266/docs/users/vector-input-ports.md)
- [Field System — framework guide](https://github.com/joby-aviation/noodles.gl/blob/58e6dfd78c5c68461dc27b8b08adec804ff85266/docs/developers/field-system.md#switchable-vector-input-ports)

## Verification

- App tests, lint, formatting, and production build
- Website production build
- Manual checks for both layouts, connection locking, direct Point connections, editable siblings, Point Lookup, persistence, and migrated projects


akre54:
Hmm. I understand the goal but I feel like this should be fixed on a type system or UI level rather than adding yet another node... Lets keep thinking on this one

chrisgervang:
The question is how do we solve a missing _input_ port issue?

Output was easy - on bbox we have multiple output types if thats what you need.

Multiple input ports that route to the same thing (position in this case) will pull conflicting values.

The user needs a way to select a another input shape

Perhaps this would be a change to MapViewState - have a way to customize the parameters from 2 lng/lat fields to 1 point field. Wdyt?

chrisgervang:
The integration repro exposed a self-dirty execution loop in the vector-channel synchronization added here.

An operator publishes a fresh vector output with `field.next()`. Parent-to-channel synchronization then called `channelField.setValue()`. Because output channel fields are owned by the producing operator, that setter immediately called `markDirty()` after the pull had made the operator clean. Point-producing operators therefore executed again on every frame.

In the reported project this affected both `/geocoder` inside a ForLoop (repeating real geocoder requests) and `/scale-container/unproject-1` on the active Deck render chain. The resulting graph churn also remounted the TableEditor schema header and closed its dialog.

Commit c7c3a677 changes only the internal parent-to-channel mirror to a non-dirtying `next()` update. User edits and channel connections still use `setValue()`. The integration repro is now stable, and the regression verifies that pulling `UnprojectOp` twice executes it only once.

chrisgervang:
@akre54 I changed directions in this PR to introduce the idea that a vec-type field can be switched from combined to channels. If you're aligned with this concept, it'd be nice to apply the feature to compound fields too.

PR #556, opened 2026-08-18
feat: Control the selected MapViewState from the map

chrisgervang:
## Summary
- enable MapLibre and pure Deck camera controls when exactly one editable MapViewState node is selected
- write longitude, latitude, zoom, pitch, and bearing back to operator fields
- guard locked, connected, multi-selected, and exporting states with clear feedback
- record a complete camera gesture as one property-history entry

## Tests
- `npm test -- --run src/noodles/utils/map-camera-control.test.ts src/noodles/utils/property-history.test.ts`
- `npm run lint`
- `npm run format:check`
- `npm run build`

### DIFF
--- a/noodles-editor/src/noodles/utils/map-camera-control.test.ts
+++ b/noodles-editor/src/noodles/utils/map-camera-control.test.ts
@@ -1,0 +1,88 @@
+import { beforeEach, describe, expect, it } from 'vitest'
+import { MapViewStateOp, NumberOp } from '../operators'
+import { getOpStore, setOp } from '../store'
+import {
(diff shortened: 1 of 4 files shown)


akre54:
Is this PR preferable to #8?

This seems more Blender-styled which I like

chrisgervang:
Yeah, I like this UX more. The fields are the same so it really doesn't warrant a new op. Interactive is just another way to change field values

#8 makes more sense if noodles were going into an interactive program direction, but for presentations I think we just need ways to bring input friction down

akre54:
Sounds good to me. Closing #8

PR #582, opened 2026-09-03
Upgrade Deck.gl to 9.4 beta 3 and use MapLibre integration

chrisgervang:
#### Background
Adopt Deck.gl 9.4 beta 3 and its dedicated MapLibre integration.

#### Change List
- Upgrade Deck.gl to `9.4.0-beta.3`, align `@luma.gl/effects` with the 9.4 shader stack, and update the lockfile.
- Replace `MapboxOverlay` with the API-compatible `MapLibreOverlay` and update the local Deck.gl alias.
- Expose the new path-style, fill-style, antialiasing, pixel-offset, view-parameter, and zoom-step properties with test coverage.


noodles-editor/src/noodles/operators.ts
```diff
@@ -4236,7 +4236,7 @@ export class DeckRendererOp extends Operator<DeckRendererOp> {
     validateViewState(viewState)
 
     // Extract geo fields from basemap so standalone DeckGL gets the correct position
-    // when basemapEnabled=false (empty mapStyle). MapboxOverlay ignores viewState, so
+    // when basemapEnabled=false (empty mapStyle). MapLibreOverlay ignores viewState, so
```
akre54, line 4239: 👍

noodles-editor/src/noodles/operators.ts
```diff
@@ -4623,12 +4625,14 @@ export class ZoomWidgetOp extends Operator<ZoomWidgetOp> {
...
   execute({
     placement,
+    zoomStep,
     viewId,
   }: ExtractProps<typeof this.inputs>): ExtractProps<typeof this.outputs> {
     const widget = {
       id: this.id,
       type: 'ZoomWidget',
       placement,
+      zoomStep,
```
akre54, line 4635: What does this give us? Isn't this an interactive-related prop?

noodles-editor/src/noodles/operators.ts
```diff
@@ -4292,6 +4292,7 @@ function createBaseViewFields() {
     }),
     clear: new BooleanField(false, { showByDefault: false }),
     clearColor: new ColorField('#00000000', { transform: hexToColor, showByDefault: false }),
+    parameters: new UnknownField({}, { showByDefault: false }),
```
akre54: I think we typically use a DataField or CompoundPropsField for parameters / properties, even when the default object is empty

noodles-editor/package.json
```diff
@@ -37,7 +37,7 @@
     "classnames": "^2.5.1",
     "d3": "^7.9.0",
     "debug": "^4.4.3",
-    "deck.gl": "^9.3.10",
+    "deck.gl": "9.4.0-beta.3",
```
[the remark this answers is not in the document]
akre54: Seems worth checking

akre54 (approved):
Nice! Seems like some positive additions

noodles-editor/src/noodles/operators.ts
chrisgervang: Yes, this is interactive. `ZoomWidgetOp` already represents the interactive zoom controls; Deck.gl 9.4 adds `zoomStep` so a project can configure the zoom-level delta per button click (for example, fractional steps). I retained it as an optional, hidden-by-default widget setting.

noodles-editor/package.json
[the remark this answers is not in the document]
chrisgervang: Valid issue. Upgraded `@luma.gl/effects` to `^9.4.0-beta.6`; `npm ls @luma.gl/effects @luma.gl/core @luma.gl/shadertools --depth=1` now reports a compatible 9.4 dependency tree.

noodles-editor/src/noodles/operators.ts
chrisgervang: Changed this to `CompoundPropsField({}, {showByDefault: false})`, matching the existing object-field convention while still allowing passthrough render parameters.

PR #584, opened 2026-09-03
Fix container rename deleting hidden children

chrisgervang:
## Summary
- route node-header renames through the canonical full-project mutation action
- preserve path-based descendants, internal edges, nested parent IDs, and operator container IDs
- remove the duplicate rename helper that operated on the visible React Flow scope

## Root cause
At the root graph level, React Flow receives only displayed root nodes and edges. The node header called its scoped setters directly, so renaming a container appeared as removal of the old container and triggered cascade deletion of children that were not in the displayed scope.

### Automated
- Focused container component and project-modification suites: 40 passed
- Full suite: 2,884 passed initially; the remaining 4 environment-sensitive tests passed when rerun with required API-key variables and without concurrent build load
- Lint and format checks passed
- Production app build passed
- Standalone TypeScript currently reports the existing repository-wide baseline backlog; no new errors were reported for the new context or rename hook implementation

### Manual test runbook
1. Open a project with a top-level container containing GraphInput, processing nodes, and GraphOutput.
2. From the parent graph, rename the container through its node header.
3. Open the renamed container.
4. Confirm every child remains present and internal and external connections point to the renamed paths.
5. Save and reload the project; confirm the graph remains intact.

## Documentation
No user-facing documentation changes needed.

### DIFF
--- a/noodles-editor/src/noodles/contexts/project-modification-actions-context.tsx
+++ b/noodles-editor/src/noodles/contexts/project-modification-actions-context.tsx
@@ -1,0 +1,38 @@
+import { createContext, type ReactNode, useContext, useMemo } from 'react'
+
+export type UpdateOperatorId = (
+  nodeId: string,
+  newBaseName: string,
+  isContainer: boolean
+) => void
+
+interface ProjectModificationActions {
+  updateOperatorId: UpdateOperatorId
+}
+
+const missingProvider = () => {
+  throw new Error('ProjectModificationActionsProvider is required for graph mutations')
+}
+
+const ProjectModificationActionsContext = createContext<ProjectModificationActions>({
+  updateOperatorId: missingProvider,
+})
+
+export function ProjectModificationActionsProvider({
+  updateOperatorId,
+  children,
+}: {
+  updateOperatorId: UpdateOperatorId
+  children: ReactNode
+}) {
+  const value = useMemo(() => ({ updateOperatorId }), [updateOperatorId])
+  return (
+    <ProjectModificationActionsContext.Provider value={value}>
+      {children}
+    </ProjectModificationActionsContext.Provider>
+  )
+}
+
+export function useProjectModificationActions() {
+  return useContext(ProjectModificationActionsContext)
+}
--- a/noodles-editor/src/noodles/noodles.tsx
+++ b/noodles-editor/src/noodles/noodles.tsx
@@ -71,6 +71,7 @@ import { SidebarTabs } from './components/sidebar-tabs'
 import { StorageErrorHandler } from './components/storage-error-handler'
 import { CanvasDropImport } from './components/tools/canvas-drop-import'
 import { UndoRedoHandler, type UndoRedoHandlerRef } from './components/UndoRedoHandler'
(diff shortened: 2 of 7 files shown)


noodles-editor/src/noodles/contexts/project-modification-actions-context.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const missingProvider = () => {
+  throw new Error('ProjectModificationActionsProvider is required for graph mutations')
+}
+
+const ProjectModificationActionsContext = createContext<ProjectModificationActions>({
+  updateOperatorId: missingProvider,
+})
+
+export function ProjectModificationActionsProvider({
```
akre54, line 21: Do we need yet another provider?

noodles-editor/src/noodles/noodles.tsx
```diff
@@ -1577,41 +1578,43 @@ export function getNoodles(): Visualization {
       >
         <PrimeReactProvider>
           <TimelineProvider>
-            <ReactFlow
```
akre54, line 1580: Hard to tell what changed here

noodles-editor/src/noodles/store.tsx
```diff
@@ -290,100 +289,6 @@ export const takePendingInsertionIndex = (
...
-//
-// Updates an operator's ID and all references to it (nodes, edges, children).
-// This is used when renaming operators in the node tree sidebar or node headers.
-//
-// @param nodeId - Current ID of the operator
-// @param newBaseName - New base name (without path prefix)
-// @param isContainer - Whether the operator is a container
-// @param setNodes - React Flow setNodes function
-// @param setEdges - React Flow setEdges function
-export const updateOperatorId = (
```
akre54, line 306: Was this duplicate code?

akre54 (approved):
Not sure we need an entire Provider for this (existing context / store should suffice) but otherwise seems fine

noodles-editor/src/noodles/noodles.tsx
chrisgervang: A provider was added

noodles-editor/src/noodles/contexts/project-modification-actions-context.tsx
chrisgervang: For now - yes. But once we make a global node/edge store and then this can get rolled into it

noodles-editor/src/noodles/store.tsx
chrisgervang: kinda - at some point a global project mutator was added. This one has been around for a while - it acted on react-flow's setNode/setEdges, which doesn't reach nodes nested in sub-graphs.

akre54: I guess the real issue is we've got multiple sources of truth then. Nodes and Edges should really only be used for ReactFlow graph rendering and should get updated from op instances rather than doing bookkeeping

chrisgervang: Yup, I expect we can remove most react-flow store code.

PR #585, opened 2026-09-03
Fix fresh-load container input pull ordering

chrisgervang:
Stacked on #561.

## Summary

- derive an ephemeral pull dependency from each external container input source to its direct GraphInput boundary
- mirror that relationship in the operator pull graph while keeping value propagation on the existing source → Container → GraphInput field path
- remove stale bridge dependencies when container input edges change
- add a deterministic regression for the fresh-load race reproduced by the LA scale container

## Why

On first load, an upstream PointOp and the container child graph could be pulled concurrently. GraphInput would then expose its empty default before PointOp published the connected value, allowing the child CodeOp to fail and remain in ERROR until another graph change (such as attaching Viewer) invalidated it.

## Validation

- npx vitest run src/noodles/container-integration.test.ts src/noodles/transform-graph.test.ts src/noodles/graph-executor.test.ts (168 passed)
- npm run lint
- npm run format:check
- npm run build

### DIFF
--- a/noodles-editor/src/noodles/container-integration.test.ts
+++ b/noodles-editor/src/noodles/container-integration.test.ts
@@ -3,7 +3,14 @@ import { afterEach, beforeEach, describe, expect, it } from 'vitest'
 import { getExecutor } from './graph-executor'
 import type { Edge } from './noodles'
 import type { OpType } from './operators'
-import { ContainerOp, GraphInputOp, GraphOutputOp, type ViewerOp } from './operators'
+import {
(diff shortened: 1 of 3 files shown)


akre54 (approved):
Seems like a lot of bookkeeping but the logic is solid. Ship it

chrisgervang:
> Seems like a lot of bookkeeping but the logic is solid. Ship it

Containers are turning into a killer feature with every polish!

PR #587, opened 2026-09-03
Provider-agnostic agent harness for the in-app chat

akre54:
## Summary

The in-app AI chat was one hardcoded provider with no bound on what entered the context window. Three concrete problems, all verified in the code before the work started:

1. **The retrieval tools existed but the chat could not call them.** `search_code`, `get_documentation`, `get_example`, `get_operator_schema`, `list_operators`, `find_symbol` and `analyze_project` were all `exposeToChat: false` and filtered out of `ClaudeClient.getTools()`. They were fully implemented and reachable from WebMCP — the in-app assistant just had no route to docs or examples. The flag was there to protect the token budget; the right fix is routing, not hiding.
2. **Tool results were unbounded.** `JSON.stringify(result)` with no cap. `listNodes` returned every node's full `inputs` *plus* three derived arrays that re-serialized the same node objects a second time. A `CodeOp`'s `code` or a `DuckDbOp`'s `sql` can be kilobytes each. On a 60-node project that is a single tens-of-thousands-of-tokens tool result — this, not the system prompt, was the actual blowup.
3. **One provider, no loop abstraction.** `new Anthropic({dangerouslyAllowBrowser: true})` with the model baked in as a static. No streaming, no cancellation (the whole multi-turn loop sat behind one spinner), no sub-agents.

This replaces all of it with one hand-rolled agent loop in `src/ai-chat/agent/` — no new dependencies, not Vercel AI SDK, not LangChain — behind a provider interface whose entire surface is a stream plus two flags. The same loop and tool surface now serve Anthropic, OpenRouter, any OpenAI-compatible endpoint, and Chrome's built-in Gemini Nano, with context cost bounded by the provider's window rather than by hand-tuned constants.

Measured on a synthetic 60-node project shaped like `nyc-taxis`: **first-turn payload drops ~48%** (~32,000 → ~16,600 chars), and a `list_nodes` result at Nano's window drops **91%** — from 18,447 chars, which exceeds Nano's entire context, to 1,637. Numbers are reproducible without an API key via `src/ai-chat/agent/context-cost.test.ts` and recorded in `dev-docs/agent-harness.md`.

## Changes

**Context hygiene**
- `agent/result-budget.ts` — `capToolResult` walks a ladder of tighter array/string limits until the payload fits, then appends a `_truncated` marker with a per-tool hint telling the model how to get the omitted detail. Truncation is element-wise, so the JSON stays valid. Budget is 10% of the window in chars, clamped 600–24,000.
- `mcp-tools.ts` `listNodes` slimmed: role groupings carry ids instead of re-serialized nodes, `position` dropped, long string inputs clipped to a preview.
- `agent/tool-router.ts` — 5 tools sent by default (`list_nodes`, `get_node_info`, `get_node_output`, `apply_modifications`, `find_tools`); `find_tools({query})` keyword-scores the whole definition list and unlocks matches for the rest of the turn. **`exposeToChat` is deleted** — it was the sole reason the chat could not reach docs.
- System prompt split into `prompts/core.md` (2,910 chars, always sent) plus six sections indexed into the docs bundle and retrieved on demand. Was 9,278 chars sent every turn.

**The harness**
- `agent/loop.ts` — steps until stop or the step limit (12, or 6 under a 32k window). Read-only tool batches run concurrently; the first mutating call forces the rest to serialize, so two `apply_modifications` can never race. Opaque `provider_block` parts ride through the transcript untouched (dropping Anthropic thinking blocks makes the next tool-result turn 400).
- `agent/subagent.ts` — `delegate({task, toolset})` runs a child loop at `depth + 1` with a fresh transcript and returns **only** its final report; the child's tool results never enter the parent transcript. Depth capped at 1.
- Compaction threshold now derives from `provider.contextWindow`, and the summarizer call goes through the provider instead of a raw `Anthropic` client.
- `claude-client.ts` deleted.

**Providers**
- `agent/providers/anthropic.ts` — keeps `cache_control: ephemeral` on the system prompt.
- `agent/providers/openrouter.ts` + `agent/providers/openai-format.ts` — shared OpenAI chat-completions wire format. The load-bearing detail: streaming tool-call `arguments` arrive as string *fragments* keyed by `index` and must be accumulated and parsed once at the end. Getting this wrong is silent tool-call corruption that looks like model failure.
- `agent/providers/custom.ts` — any OpenAI-compatible endpoint (Groq, OpenAI, vLLM, LM Studio). `validateCustomEndpoint` probes `GET {base}/models` on Save and refuses a server whose catalogue omits the model you typed. A browser "Failed to fetch" is reported as "connection refused, or the server does not allow browser requests (CORS)", which is what it almost always is.
- `agent/providers/chrome.ts` — no tool calling, so the loop runs constrained-JSON action mode over the router-narrowed tool list and translates the one-object reply into the same `tool_call` event everything else emits. Download progress, 10-minute download timeout, `chrome://flags` hint on failure.
- `agent/web-search.ts` — server-side tool on Anthropic, `plugins: [{id:'web'}]` on OpenRouter, not offered at all on Chrome or custom endpoints so the model never promises a search it cannot run.

**UI and keys**
- Streaming text, collapsible tool rows, live token+cost readout from the `usage` event, a Stop button wired to the `AbortController`, provider and model selects in the header.
- `providerPreference` in `keys-store.tsx` is the single source of provider selection; `'automatic'` walks anthropic → openrouter → custom → chrome and takes the first with a credential. Legacy `'chrome-ai'` reads as `'chrome'`.
- `openrouter` added to `KeyType`, the `setBrowserKeys` allowlist, and `getEnvKeys` as `VITE_OPENROUTER_API_KEY`.
- Transcript fidelity: `Message` gained `toolUses: {name, params, ok}[]`, replayed as one text line. Deliberately not full `tool_use`/`tool_result` replay — that would put every past tool result back into context on every later turn, which is the cost this PR exists to bound. Stored conversations are `version: 2`; version 1 records still load.

**Relationship to #576.** #576 landed a second provider layer on main while this branch was replacing the first one. The two shapes cannot coexist: `ai-chat/providers/` gives each provider its own conversation loop, compaction threshold and copy of the system prompt; `ai-chat/agent/` has one loop and asks a provider only for a stream. After the rebase main's copies no longer compiled (they import the deleted prompt file, filter on the removed `exposeToChat`, and call the old `compactConversation` signature) and nothing imported them. The last commit deletes `ai-chat/providers/` and ports every capability #576 added — custom endpoints (now with streaming, which #576 lacked), the `/models` validation (now stricter), Chrome download progress, the provider-error screen, the settings deep-link, and the AI Provider settings tab kept as main wrote it. Not ported: #576's bespoke provider dropdown and context-window donut, since the header already has both selects and a live cost readout.

It also fixes a bug that came in with #576 and is live on main today: `settings-dialog.tsx` referenced `removeProjectKey` without ever reading it from the store, so the project-key Remove button threw a `ReferenceError`.

### Unit Tests

`npx vitest run` — **156 files, 3,044 passed, 10 skipped, 0 failed.**

New suites:
- `result-budget.test.ts` — a 100-node synthetic project stays under budget; truncation yields valid JSON with a `_truncated` marker; long `CodeOp.code` is clipped.
- `tool-router.test.ts` — the default set is exactly 5 tools; `find_tools('documentation')` surfaces `get_documentation`; a Nano-sized window never widens past the default set.
- `loop.test.ts` — with a fake provider yielding scripted events: multi-step loops terminate, the step limit is enforced, read-only batches run in parallel, a mutating call forces serialization, abort mid-stream cleans up.
- `subagent.test.ts` — the parent transcript contains only the child's summary, never its tool results; `depth > 1` is refused.
- `openai-format.test.ts` — SSE tool-call fragment reassembly against a captured chunked stream, at several chunk boundaries.
- `custom.test.ts` — base-URL normalization, `chat/completions` request shape, `stream_options` vs `usage`, tools omitted when unsupported, CORS/unreachable messaging, endpoint validation.
- `chrome.test.ts` — constrained-JSON parsing.
- `context-cost.test.ts` — the before/after payload figures above, so the claim is checkable rather than asserted.

Existing suites stay green, including `tool-definitions.test.ts` (updated for the removed field), `conversation-history.test.ts` (version 1 records still load) and `webmcp/webmcp.test.ts` (every tool still registers). No test spends a real API request.

### Manual Testing

Assumes the app is running at `http://localhost:5173/examples/nyc-taxis`. Open the chat panel; provider and model selects are in its header, keys in Settings → API Keys, provider pinning in Settings → AI Provider.

1. **Anthropic, streaming and abort.** With only an Anthropic key set, ask *"why is my scatterplot not showing?"* → text streams in token by token, tool rows appear as they run, and the Stop button ends the turn mid-stream without leaving a spinner.
2. **OpenRouter, cost readout.** Add an OpenRouter key, pick `google/gemini-2.5-flash`, ask the same question → identical tool behaviour, and the header shows a live token count and dollar cost from the `usage` event.
3. **Progressive disclosure.** Ask *"how do I load a CSV?"* → the tool rows show `find_tools` firing first, then `get_documentation`. **This path is impossible on main** and is the clearest proof the routing works.
4. **Sub-agents.** Ask *"find an example that animates arcs"* → a `delegate` row appears, and the answer arrives without the child's `search_code` hits landing in the transcript.
5. **Custom endpoint.** Settings → AI Provider → Custom, enter `http://localhost:1234/v1` with an LM Studio model name and press **Test and Save** → it validates against `/models`. Deliberately mistype the model name → Save is refused with the available models listed. Point it at a dead port → the error names the endpoint and mentions CORS rather than surfacing "Failed to fetch".
6. **Chrome built-in.** In Chrome Canary with the Prompt API enabled, pin Chrome and ask *"what nodes are in this project?"* → the download progress bar shows a real percentage on first use, then a `list_nodes` → answer turn completes without `QuotaExceededError`. Without the flag, the gate screen names the provider and links `chrome://flags/#prompt-api-for-gemini-nano`.
7. **WebMCP is untouched.** Open with `?externalControl=true` → `register.ts` still logs the full tool count. The chat's routing is a chat-side concern; external clients keep the whole surface.

## Documentation

- New `dev-docs/agent-harness.md` — layout, the provider table, the loop, routing, budgets, harness tools, transcript fidelity, and the measured before/after context cost with the command to re-derive it.
- `AGENTS.md` — the AI-assistant section rewritten around the harness: the five things to know before changing it.
- `dev-docs/webllm-ai-integration.md` and `dev-docs/specs/webllm-ai-chat/webllm-ai-chat.md` marked superseded. Those 4,500 lines were written in PR #126, propose LangChain.js + `@mlc-ai/web-llm` + `voy`, and were never implemented — no matching code exists.

## Related Issues

Supersedes the provider layer added in #576.


akre54:
Not perfect but it's a start. Merging

PR #595, opened 2026-09-05
Fix TableEditor schema persistence and defaults

### LINKED ISSUE #594: TableEditor loses inherited schemas and ignores declared defaults
TableEditor schemas shared through connections can serialize as null columns after the schema edge is disconnected, leaving rows empty after reload. Missing string-literal and vector cells also render generic fallbacks instead of the schema defaults, and saving an unchanged schema can appear to reset custom columns.

Expected behavior: a disconnected TableEditor retains and serializes the inherited schema, materializes declared defaults, and preserves compatible custom values on schema save.

chrisgervang:
Closes #594

#### Background
TableEditor nodes could lose shared inherited schemas during serialization, render generic fallbacks instead of declared defaults, and reset custom cells during schema saves.

Previously, a schema connection that was later disconnected serialized columns to null:
Screenshot 2026-09-05 at 8 57 51 AM

#### Change List
- Preserve repeated non-circular values during project serialization so disconnected schemas persist.
- Normalize TableEditor data against connected and edited schemas while preserving compatible custom values.
- Add regression coverage for schema disconnects, string-literal and vector defaults, and unchanged schema saves.

### DIFF
--- a/noodles-editor/src/noodles/components/table-editor.test.tsx
+++ b/noodles-editor/src/noodles/components/table-editor.test.tsx
@@ -247,6 +247,84 @@ describe('TableEditor', () => {
     expect(getByText(/1 row × 2 columns/i)).toBeDefined()
   })
 
+  it('applies declared defaults when an inherited schema adds columns', () => {
+    const onDataChange = vi.fn()
+    const onSchemaChange = vi.fn()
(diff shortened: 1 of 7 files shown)


chrisgervang:
@greptileai

PR #598, opened 2026-09-05
Fix optional compound property connection validation

### LINKED ISSUE #597: Compound connection validator rejects omitted optional properties
The structural connection validator treats every target object property as required, even when its Zod schema accepts `undefined`. This produces a false `compound cannot connect to compound` error for `BoundingBoxOp.viewState` connected to `MaplibreBasemapOp.viewState`, although the connection executes correctly. Update compatibility checking to allow omitted optional target properties while continuing to validate their types when present, and add regression coverage for the operator pair.

chrisgervang:
Closes #597

#### Background

Compound connections incorrectly reported a type mismatch when optional target properties were absent from the source.

Before
False compound type mismatch before the fix

After
Valid compound connection after the fix

#### Genuine type mismatch

When schemas are genuinely incompatible, the error now identifies the target property and expected and received types.

Improved error message for a genuine compound type mismatch

Reproduction project: [download `noodles.json`](https://github.com/user-attachments/assets/06a9f238-5a16-4fe6-9863-9462cecac689). The attachment API requires a media filename, so save the unchanged payload as `noodles.json` before opening it in Noodles.

#### Change List
- Allow omitted target properties when their schemas accept `undefined`, while still validating supplied values.
- Report the first incompatible target property with runtime type details when available.
- Add structural and `BoundingBoxOp` to `MaplibreBasemapOp` regression tests.
- Verified with 339 focused tests, lint, formatting, and a production build.

### DIFF
--- a/noodles-editor/src/noodles/operators.test.ts
+++ b/noodles-editor/src/noodles/operators.test.ts
@@ -69,6 +69,7 @@ import {
 } from './operators'
 import { deleteOp, getOpStore, setOp } from './store'
 import { isAccessor } from './utils/accessor-helpers'
+import { canConnect } from './utils/can-connect'
 
 describe('basic Operators', () => {
   it('creates an Operator', () => {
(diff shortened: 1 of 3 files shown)


chrisgervang:
@greptileai please review the latest commit.

PR #607, opened 2026-09-05
Persist edits to connected TableEditor data

chrisgervang:
## Summary

- keep converted TableEditors live to their connected source until the first edit
- persist connected-table edits as a local data override without weakening generic connected-input serialization
- restore the override on reload and retain empty-table edits
- keep unconnected TableEditor serialization unchanged

## Regression

Viewer-to-TableEditor conversion preserves the incoming data edge. Generic project serialization and property-history snapshots correctly omit connected input values, so edits made in converted tables appeared locally but were discarded when saving.

## Test plan

- `npm test -- --run src/noodles/operators.test.ts src/noodles/utils/serialization.test.ts src/noodles/components/table-editor.test.tsx` (377 passed)
- `npm run lint -- --diagnostic-level=error`
- `npm run build`

Manually: convert or open a TableEditor connected to a FileOp, edit a cell, save/reload, and confirm the edit remains while the source edge is preserved.

### DIFF
--- a/noodles-editor/src/noodles/components/op-components.tsx
+++ b/noodles-editor/src/noodles/components/op-components.tsx
@@ -1776,7 +1776,7 @@ export function TableEditorOpComponent({
   type,
   selected,
 }: ReactFlowNodeProps<NodeDataJSON<TableEditorOp>> & { type: 'TableEditorOp' }) {
-  const op = getOp(id as string)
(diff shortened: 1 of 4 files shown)


chrisgervang:
Closing because the affected TableEditors are already disconnected standalone nodes. A hidden connected-data override is the wrong persistence model; investigating the actual standalone serialization regression instead.

PR #480, opened 2026-05-29
feat: add one-click ViewerOp to TableEditorOp conversion

akre54:
#### Background

Users frequently start by inspecting data with ViewerOp, then want to edit it with proper column types. Previously, this required manually creating a TableEditorOp, connecting it, and configuring the schema. This PR adds a simple one-click conversion button that streamlines this workflow.

#### Change List
- Added `convertViewerToTableEditor()` utility function that handles operator replacement with automatic schema inference
- Added "Convert to Table Editor" button to ViewerOp component that appears when displaying tabular data (arrays of plain objects)
- Conversion preserves position, connections, and locked state with full undo/redo support
- Added comprehensive unit tests (6 tests covering success and failure cases)
- Fixed import ordering in test files (linter cleanup)

PR #423, opened 2026-04-26
feat: TableEditorOp v2 with typed schema system

akre54:
# TableEditorOp v2: Typed Schema System

This PR introduces a complete rewrite of the TableEditorOp with a typed schema system, providing an Airtable-like experience with Noodles-native field types.

## ✅ Status: Ready for Review

All features implemented and tested. The runtime integration issue has been **resolved**.

### 1. Typed Column System
- **10 column types** with specialized editors:
  - `number` - Scrubbable NumberInput with min/max/step
  - `string` - Text input
  - `boolean` - Checkbox toggle
  - `color` - Color picker
  - `point2d` - Dual number inputs [lng, lat] with geocoder button
  - `point3d`, `vec2`, `vec3` - Vector inputs
  - `date` - Date picker
  - `stringLiteral` - Dropdown for enum values

### 2. Schema Editor Dialog
- Add, edit, delete, and reorder columns
- Set column types and options
- Quick-add templates:
  - Position XYZ (3 number columns)
  - Color RGB (3 number columns 0-255)
  - Lat/Lng (point2d column with geocoder)

### 3. Hybrid Schema Approach
- **Infers schema from data** when schema input not provided
- **Allows explicit schema definition** for empty tables
- **Validates data** against schema on every execution
- Backward compatible with existing projects

### 4. TanStack Table Integration
- Headless architecture (~25KB)
- No stale closure bugs (fixed issues from PRs #374 and #286)
- Type-safe column definitions
- Easy to extend for future keyframing support

### New Files
- `src/noodles/table-schema.ts` - Core schema system (340 lines, 28/28 tests passing)
- `src/noodles/components/table-editor-v2.tsx` - TanStack Table component (580 lines)
- `src/noodles/components/schema-editor-dialog.tsx` - Schema management UI (280 lines)
- `src/noodles/table-schema.test.ts` - Comprehensive unit tests
- `src/noodles/components/table-editor-v2.test.tsx` - Component tests
- `docs/table-editor-v2.md` - Complete documentation (600 lines)
- `public/noodles/table-editor-demo.json` - Example project

### Modified Files
- `src/noodles/operators.ts` - Updated TableEditorOp class (simple pass-through)
- `src/noodles/components/op-components.tsx` - New TableEditorOpComponent
- `package.json` - Added @tanstack/react-table dependency

### Bug Fixes
- **Runtime error resolved**: Replaced CommonJS `require()` with ES6 imports to fix "ReferenceError: require is not defined" in browser
[...]

PR #608, opened 2026-09-06
Show TableEditor data connections

chrisgervang:
## Summary

Restore the TableEditor `par.data` input row and React Flow handle so preserved Viewer connections remain visible and removable after conversion. The editable table remains rendered while the input is connected.

## Historical intent

PR #423 intentionally made the table itself the typed editor and hid the redundant generic `data` row, but did not describe hiding live connections. PR #480 later explicitly preserved Viewer connections during conversion. The invisible edge was an accidental interaction between those choices; #605's duplicate-edge repair is valid but unrelated.

## Changes

- Render all visible TableEditor inputs, including `data`.
- Add a connected-component regression test that verifies both the `par.data` handle and upstream rows are rendered.

### Unit Tests

- `npm test -- --run src/noodles/components/__tests__/table-editor-op-component.test.tsx`
- `npx biome check src/noodles/components/__tests__/table-editor-op-component.test.tsx`

The repository-wide TypeScript check currently reports existing baseline errors unrelated to this two-line behavior change.

### Manual Testing

1. Connect a tabular data output to a Viewer.
2. Convert the Viewer to a Table Editor.
3. Confirm the incoming data edge and `data` field row remain visible.
4. Confirm the editable table remains visible and contains the upstream rows.
5. Select and delete the incoming edge; confirm the Table Editor remains usable as a standalone table.

## Documentation

No documentation changes needed.

## Related Issues

This fixes the invisible-connection interaction discovered while testing #605, but is independent of its duplicate-edge behavior.

### DIFF
--- a/noodles-editor/src/noodles/components/__tests__/table-editor-op-component.test.tsx
+++ b/noodles-editor/src/noodles/components/__tests__/table-editor-op-component.test.tsx
@@ -1,0 +1,72 @@
+import { cleanup, render, screen } from '@testing-library/react'
+import { ReactFlowProvider } from '@xyflow/react'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { ProjectModificationActionsProvider } from '../../contexts/project-modification-actions-context'
+import type { TableEditorOp } from '../../operators'
+import { clearOps, getOp } from '../../store'
+import { transformGraph } from '../../transform-graph'
+import { nodeComponents } from '../op-components'
+
+describe('TableEditorOpComponent', () => {
+  beforeEach(() => {
+    clearOps()
+  })
+
+  afterEach(() => {
+    cleanup()
+    clearOps()
+  })
+
+  it('renders its data input handle without hiding the table editor', () => {
+    const schema = {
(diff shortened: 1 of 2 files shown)


chrisgervang:
Having invisible edges is problematic. Did you mean to disconnect them in #480, or make the data port visible again?

akre54 (changes requested):
Sorry what's the use case? I purposefully removed the input so that we could remove state from the graph. 

If #480 neglected to cleanup the subscriptions that's an oversight

chrisgervang:
#480 did not remove the edges, so the subscriptions remained.

They should be removed during viewer to table conversion if they're not meant to update.

If you have the time, it'd help me if you could write down the spec you've intended for this op - design docs are sparse, so I'm needing to re-interpret the intent from clues whenever there's buggy behavior

akre54:
Got it thanks. Yeah I'll try to remember to document where I can. The spec is evolving pretty drastically and it's less formalized than I'd like it to be (we don't have PMF). But I'll try to take a crack at it

akre54:
#611

Issue #122, opened 2025-11-05
[Feat] Locking behavior updates


akre54:
### Target Use Case

Our locking feature is based on TouchDesigner's - it preserves changes and ignores new data. Some of the functionality can be made more obvious or we can improve the UX. A few features based on convo with @chrisgervang 

### Proposal

In order of importance and difficulty:

1. When a user unlocks an op, we should have them confirm (via a modal or js `confirm` popup) that they understand that this will overwrite the data if there are any incoming edges
2. Style edges leading to locked ops as grayed out to make it obvious the change has no effect
3. Add a cache folder with serialized cache files in the project directory. Each cache file should be named by op id (make sure to update the filename if the id changes). This should ensure state is reconstructed on load without bloating the noodles.json file
4. Add a history entry to the undo for unlocking ops. Maybe even batch multiple TableEditorOp edits together into a single history entry

PR #609, opened 2026-09-06
RFC: merge connected TableEditor edits by row key

chrisgervang:
## Summary

Draft an RFC for preserving local TableEditor edits while an upstream `data` connection remains live.

The proposal keeps standalone tables unchanged and replaces the full-table override idea with a stable-keyed `add`/`update`/`delete` change set plus cell-level three-way reconciliation.

## Proposal

- Require explicit stable or composite row keys for connected editing.
- Store only touched cells, inserted rows, and deleted-row preimages.
- Compare base, local, and latest-upstream values to merge unrelated changes and surface true conflicts.
- Preserve lock as the separate snapshot/freeze workflow.
- Materialize the reconciled table into ordinary `inputs.data` when disconnecting.
- Avoid array-index JSON Patch and whole-array JSON Merge Patch semantics.

## Research

The RFC compares the design with:

- AG Grid row transactions and stable row IDs
- Entity Framework optimistic concurrency resolution
- Kubernetes keyed list merging and field ownership
- Delta Lake preimage/postimage change records
- PostgreSQL conflict keys and upserts
- JSON Patch and JSON Merge Patch limitations for tabular arrays

## Testing

Documentation-only draft; no runtime behavior changes.

## Review questions

1. Should key metadata live in `TableSchema` or a dedicated input?
2. Should inserts require complete keys before leaving the row editor?
3. Should data-source operators advertise primary-key metadata?
4. How should schema renames remap existing edits?

## Related work

- #608 restores the visible/removable TableEditor data connection independently.
- #122 documents the existing snapshot-style lock intent.
- #423 introduced the typed editable table.
- #480 introduced connection-preserving Viewer conversion.

### DIFF
--- a/dev-docs/specs/table-editor-row-deltas.md
+++ b/dev-docs/specs/table-editor-row-deltas.md
@@ -1,0 +1,342 @@
+# RFC: Mergeable TableEditor row deltas
+
+Status: Draft for discussion
+
+## Decision summary
+
+A standalone `TableEditorOp` should continue to own and serialize its complete table in
+`inputs.data`.
+
+An unlocked, connected `TableEditorOp` should not save a second full table. It should save a
+small, row-keyed change set containing inserts, deletes, and the cells changed by the user. On
+each upstream update, the operator should deterministically apply that change set and identify
+conflicts by comparing the original upstream value, the user's value, and the latest upstream
+value.
+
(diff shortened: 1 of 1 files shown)


akre54 (commented):
My preference is to leave the table editor alone.

If there's a secondary operator that applies these changes as patches that feels like a different operation

Can you tell me more about the use case?

